PR #274, opened 2014-12-16
Remove encoding cookies from 2 Casks

rolandwalker:
These UTF-8 encoding cookies were intended to help Ruby 1.9, and should not have had an effect for users running Ruby 2.0.  We no longer need them.

Since caskroom/homebrew-cask#8089 (requiring Ruby 2.0) was only just merged, we should HOLD until 15 Jan.

Sorry if this long-term PR is considered clutter.


ndr-qef:
Following from caskroom/homebrew-cask#8154.

PR #295, opened 2015-01-27
Updated EB Garamond with real upstream source and newer version.

pal:
This was previously just an outdated Google Web Font source with just regular, now from proper source and the latest version.


rolandwalker:
Yes, Google Web Fonts were a great way to seed the repo, but Google usually lags the upstream.  There are actually hundreds here that could be updated as you have done for this one.

Many thanks!

PR #297, opened 2015-01-30
Updating Font Awesome Cask

andres-torres-marroquin:
The current font-fontawesome cask is broken, because the zip file is not longer available, I updated the sha and the download link for the version 4.2.0

### DIFF
--- a/Casks/font-fontawesome.rb
+++ b/Casks/font-fontawesome.rb
@@ -1,8 +1,8 @@
 cask :v1 => 'font-fontawesome' do
   version '4.2.0'
-  sha256 'b0f2792a2a096ccebe72cfcf1880671658a1f18d5356c5f991b2b54d8f6bf704'
+  sha256 'b68b47c5e6da744fc78393eeae0cb16e3844a76b505855c99ff50e99c680b5c0'
 
-  url "http://fontawesome.io/assets/font-awesome-#{version}.zip"
+  url "https://github.com/FortAwesome/Font-Awesome/archive/v4.2.0.zip"
   homepage 'http://fortawesome.github.io/Font-Awesome/'
   license :oss
 


rolandwalker:
Thanks!

Amorymeltzer:
The cask was "broken" because the version was outdated - it's now 4.3.0.  If the owner is going to remove links once they become outdated we might as well use the [releases page on github](https://github.com/FortAwesome/Font-Awesome/releases) I agree, but even then, shouldn't the url be `https://github.com/FortAwesome/Font-Awesome/archive/v#{version}.zip`?

Amorymeltzer:
The cask was "broken" because the version was outdated - it's now 4.3.0.  If the owner is going to remove links once they become outdated we might as well use the [releases page on github](https://github.com/FortAwesome/Font-Awesome/releases) I agree, but even then, shouldn't the url be `https://github.com/FortAwesome/Font-Awesome/archive/v#{version}.zip`?

I'll also note that the sha appears to differ between the download on the releases page and that on the homepage.

rolandwalker:
@Amorymeltzer then yes the release page on GitHub looks like a better source.  I'll look at this again tomorrow if someone else doesn't get to it first.

It's too bad Travis doesn't check the SHA for the Casks changed by a given PR.  It seems like it would be possible to hack something together to do that.

As to re-use of the version number, it is nice to have, but I'm strongly against blocking a submission that works and contains nothing incorrect.  That extends to other lint-y details such as
 * whitespace
 * layout/order
 * license value of `:unknown`

We can always catch up with those items later.  It makes a nicer experience for our very generous contributors if we avoid quibbling.

Amorymeltzer:
@rolandwalker :+1: Just checkin' my sanity.

PR #303, opened 2015-02-16
Update font-comic-neue.rb

nakp:
I'm still not sure if I should keep ttf or move to otf


rolandwalker:
Thanks!  Switching to the OTF seems reasonable.

PR #314, opened 2015-04-05
Update font-everson-mono to version 7.0.0 from 6.2.1

treyharris:
Unfortunately version number is encoded in zipfile.


ndr-qef:
Thanks.

PR #321, opened 2015-05-03
Fix CJK font file name

unasuke:
CJK fonts
(Japanese, Korean, Simplified Chinese, Traditional Chinese)
These fonts file name are different from other Noto font file name.
This commit is fix that.


ndr-qef:
Not checking the filenames after the recent URL change was an oversight on my part. Thanks.

PR #325, opened 2015-05-15
Update url for Google fonts which moved from apache to ofl.

rockymadden:
I had identified `exo2` had an issue and did a search through out the whole of homebrew-fonts with a matching pattern for `trunk/apache` in the url. I checked to see if each one had moved in the google-fonts project and if they had I updated the url.


ndr-qef:
Thanks, this is really helpful.

PR #328, opened 2015-05-19
Add cask for Fira Code v0.3

martinbmadsen:
The Firefox Fira monospaced font with programming ligatures, based on Hasklig.

### DIFF
--- a/font-fira-code.rb
+++ b/font-fira-code.rb
@@ -1,0 +1,10 @@
+cask :v1 => 'font-fira-code' do
+  version '0.3'
+  sha256 'a90721b6b1de4640ace744657432a990aae4c8e403b78141a50d0a7abd305db7'
+
+  url 'https://github.com/tonsky/FiraCode/releases/download/0.3/FiraCode-Regular.otf'
+  homepage 'https://github.com/tonsky/FiraCode'
+  license :ofl
+
+  font 'FiraCode-Regular.otf'
+end


font-fira-code.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'font-fira-code' do
+  version '0.3'
+  sha256 :no_check
```
vitorgalvao: `sha256` also needs to be updated. You can find what it is by running `shasum -a 256` on the downloaded file.

font-fira-code.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  url 'https://github.com/tonsky/FiraCode/releases/download/0.3/FiraCode-Regular.otf'
+  homepage 'https://github.com/tonsky/FiraCode'
+  license :ofl
+
+  font 'FiraCode-Regular.otf'
+
```
vitorgalvao: Please remove this empty line.

martinbmadsen:
Updated to have sha256... my bad. I was using https://github.com/caskroom/homebrew-fonts/blob/master/CONTRIBUTING.md as an example, and noticed that the font-inconsolata didn't have a sha256 check because it was not a container, so I thought the same was true for this font.

But anyway, updated now!

vitorgalvao:
> noticed that the font-inconsolata didn't have a sha256 check because it was not a container

It has nothing to do with that. It doesn’t have a checksum because it doesn’t have a version, which means it is updated in place, and the cask would then break constantly.

PR #338, opened 2015-06-04
Add XITS-math font

martinbjeldbak:


ndr-qef:
Thanks. Before merging, we need a small correction: according to the OTF files, the family name is `xits` rather than `xits-math`. Could you amend the commit?

martinbjeldbak:
Makes sense! My bad... Done now. :+1:

ndr-qef:
Excellent, thank you.

PR #354, opened 2015-07-03
Add Palemonas font

tghs:
Details in English: http://www.mif.vu.lt/~vladas/Palemonas.htm

### DIFF
--- a/Casks/font-palemonas.rb
+++ b/Casks/font-palemonas.rb
@@ -1,0 +1,14 @@
+cask :v1 => 'font-palemonas' do
+  version '3.0'
+  sha256 'ad576fc82d1f67824cc10cde6c1778016ecf16702bcc2353f0d3884f2283a978'
+
+  url 'http://www.vlkk.lt/media/public/file/Palemonas/Palemonas-3_0.zip'
+  homepage 'http://www.vlkk.lt/palemonas'
+  license :gratis
+  container :type => :zip
+
+  font "Palemonas-#{version}/Palemonas-nm.ttf"
+  font "Palemonas-#{version}/Palemonas-bd.ttf"
+  font "Palemonas-#{version}/Palemonas-it.ttf"
+  font "Palemonas-#{version}/Palemonas-bi.ttf"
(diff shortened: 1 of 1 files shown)


Casks/font-palemonas.rb
```diff
@@ -0,0 +1,14 @@
+cask :v1 => 'font-palemonas' do
+  version '3.0'
+  sha256 'ad576fc82d1f67824cc10cde6c1778016ecf16702bcc2353f0d3884f2283a978'
+
+  url 'http://www.vlkk.lt/media/public/file/Palemonas/Palemonas-3_0.zip'
+  homepage 'http://www.vlkk.lt/palemonas'
+  license :gratis
+  container :type => :zip
```
vitorgalvao: Are you sure this is needed? Shouldn’t be.

Casks/font-palemonas.rb
```diff
@@ -0,0 +1,14 @@
+cask :v1 => 'font-palemonas' do
+  version '3.0'
+  sha256 'ad576fc82d1f67824cc10cde6c1778016ecf16702bcc2353f0d3884f2283a978'
+
+  url 'http://www.vlkk.lt/media/public/file/Palemonas/Palemonas-3_0.zip'
```
vitorgalvao: `url "http://www.vlkk.lt/media/public/file/Palemonas/Palemonas-#{version.sub('.','_')}.zip"`.

tghs: Thanks, I'll resubmit with that change squashed in.

Casks/font-palemonas.rb
tghs: Yes, I initially tried without that. But then you get this error:

```
Error: Uh oh, could not identify primary container for '/Library/Caches/Homebrew/font-palemonas-3.0.zip'
```

I've not dug any further into why the primary container detection is getting tripped up by this file.

vitorgalvao: Sounds good. Leave it, then.

PR #364, opened 2015-07-24
Add Base Monoid font

sscotth:
Not sure how best to handle all the variants: https://github.com/larsenwork/monoid/tree/release


vitorgalvao:
[A cask is needed for each one](https://github.com/caskroom/homebrew-cask/pull/1999).

sscotth:
Created a cask for all variants.

vitorgalvao:
Holy crap. Are you telling me there are 3200 variants? That is insane.

They *all* need a `# github.com is the official download host per the vendor homepage` comment above `url` ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5)). However, even though all your work is appreciated, I’m very wary of merging this. As soon as their release strategy changes (which could very well happen, especially since they’re relatively recent) we’ll have over 3000 files to deal with, and I really don’t feel like being the one to deal with that. [This tweet](https://twitter.com/ncoghlan_dev/status/598146064508538880) comes to mind.

For context on how insane 3000 files for a single font is, we have under 1000 font casks, and only 2651 casks in the main repo.

Pinging @ndr-qef since you’ve been dealing with this repo more. I’d also like to know if @jawshooah and @Ngrd have an opinion on the matter. I say we can’t accept this, and the best course of action is for someone to provide and maintain a tap. I don’t even find it easy to believe anyone who would want to install this via homebrew-cask would go through the trouble of understanding exactly what cask they need.

In case you’re wondering how they manage all this for people who want to download them, they provide a way to select what you want on the website, prior to downloading.
image

Ngrd:
This is truly insane to merge this in my opinion. @sscotth, I really do appreciate all the hard work you have done to prepare this PR. You are awesome! But for `homebrew-fonts` this would totally be insane.

I fully agree with @vitorgalvao, probably the best option would be to make a separate tap for this, but I can't imagine who will be able to maintain this all the time. I don't say, that this is impossible, and most likely the original repository for Monoid font won't change very often when it becomes more stable. However, having a separate tap only for that also seems like a bizarre idea.

P.S. This PR is truly epic.

jawshooah:
Agreed with @vitorgalvao and @Ngrd; it just isn't reasonable to maintain this many files for a single font.

The best solution in my view would be to select one version as the "default" and include it here. If someone else, perhaps @sscotth, feels inclined to create and maintain a custom tap with the thousands of other variants, they are free to do so.

vitorgalvao:
Seems like we have some consensus. I’d even be OK with including in the cask a `caveat` stating

```
This cask only installs the Normal weight with medium line height.
To get other styles, please tap the <whatever> repo
  brew tap <whatever>
```

That way whoever installs the font also knows where to look for other styles.

What do you think, @sscotth? If you need help [creating a tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md) we can help. It isn’t hard.

sscotth:
It wasn't bad, wrote a simple node script to generate the casks.

Adding the Github comment is no problem, I just noticed many fonts that didn't have such a mention. https://github.com/caskroom/homebrew-fonts/blob/master/Casks/font-abeezee.rb

Should I just add the tap comment above as a comment in the cask?

vitorgalvao:
> I just noticed many fonts that didn't have such a mention.

That’s bad, and we need to fix those. Keep in mind that is only necessary when the host of `homepage` and `url` are different.

> Should I just add the tap comment above as a comment in the cask?

Not a comment, but a [caveat](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#caveats-stanza-details) (https://github.com/caskroom/homebrew-cask/blob/458a64e939a8648f990e81a8c8aa95afa7899639/Casks/world-of-tanks.rb#L17-21), so something like

```ruby
caveats <<-EOS.undent
#{token} only installs the Normal weight with medium line height.
To get other styles, please tap the <whatever> repo
  brew tap <whatever>
EOS
```

This ensures people will be informed when `installing` and `info`ing the cask.

gtramontina:
@sscotth @vitorgalvao  @larsenwork Thank you all for the effort!
Any thoughts on getting https://github.com/sscotth/homebrew-fonts/commit/68fdead4c09fdde2b5a448c8177494c270978f83 amended with the caveat and merging it? I mean, anytime soon?

PR #366, opened 2015-07-28
update Amiri source to latest version and real upstream

alerque:


Casks/font-amiri.rb
```diff
@@ -1,17 +1,13 @@
...
-  sha256 :no_check
+  version '0.107'
+  sha256 '7e81fa2844ded0d7839dc8eedcc4f7ece3688b9cd9ac9b3fd806340cdcef5463'
 
-  url 'https://github.com/google/fonts/trunk/ofl/amiri',
-      :using      => :svn,
-      :revision   => '50',
-      :trust_cert => true
-  homepage 'https://www.google.com/fonts/earlyaccess'
+  url 'https://github.com/khaledhosny/amiri-font/releases/download/0.107/amiri-0.107.zip'
```
vitorgalvao: Needs a `# github.com is the official download host per the vendor homepage` comment above this line ([example](https://github.com/caskroom/homebrew-cask/blob/cb487ad405f74cb1f26b93e3f3e8412c82a22e34/Casks/0ad.rb#L5)).

Also, please add a `appcast 'https://github.com/khaledhosny/amiri-font/releases´' line under.

Please @mention me after you do, so I get an email to check back.

alerque:
@vitorgalvao Done. I assume (based on the only other example I see in the repo) that the appcast URL is actually supposed to be the feed not the page as you had linked, was that correct?

vitorgalvao:
Yes, you are correct. Copied the link and forgot to add the `atom` part. Now that I thing about it, I should’ve already made a TextExpander shortcut for this.

PR #369, opened 2015-07-30
Remove Envy Code R

scw:
`brew cask audit` showed this cask as missing a license, but when I
went looking for one, I found this instead:

> These files are free to download and use from damieng.com but CAN
> NOT be redistributed either by other web sites or be included in
> your package, download, product or source repository at this time.

Based on the author's wishes, this font should probably be removed.


damieng:
Hi, author of Envy Code R here.

Only just come across this - this kind of use where it downloads the font directly from damieng.com is completely fine.

vitorgalvao:
Good to know. Thank you for stopping by and clarifying, @damieng.

@scw Want to resubmit with the fixes for the changes introduced to homebrew-cask since then?

PR #371, opened 2015-08-01
Add Monoisome font

sscotth:
New font-family from Monoid.

> Monoid + Font Awesome = Monoisome. See the icons directly in your code.


sscotth:
This may require a caveat in the future, and I will update as I track progress with the sscotth/homebrew-monoid tap.

PR #396, opened 2015-10-05
Add the Octicons font

benbalter:
This adds GitHub's Octicons font to Homebrew, literally, the only font I use which isn't already part of the cask.

To note, it will install off of `master`, not truly the latest release.

Fixes https://github.com/github/octicons/issues/70. 

/cc @cameronmcefee

### DIFF
--- a/Casks/font-octicons.rb
+++ b/Casks/font-octicons.rb
@@ -1,0 +1,10 @@
+cask :v1 => 'font-octicons' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://github.com/github/octicons/archive/master.zip'
+  homepage 'https://octicons.github.com'
+  license :ofl
+
+  font 'octicons-master/octicons/octicons-local.ttf'
+end


Casks/font-octicons.rb
```diff
@@ -0,0 +1,11 @@
+cask :v1 => 'font-octicons' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://github.com/github/octicons/archive/master.zip'
+  homepage 'https://octicons.github.com'
+  caveats "When using the GitHub logos, be sure to follow the GitHub logo guidelines (https://github.com/logos)"
```
vitorgalvao: This does not concern us. We’re not a discoverability service, so you should already know what you’re getting. `caveats` are for clarifications with installations, not usage. Please remove it.

benbalter: Thanks for the :eyes:. Removed via 65b79e1.

Casks/font-octicons.rb
```diff
@@ -0,0 +1,10 @@
+cask :v1 => 'font-octicons' do
+  version :latest
+  sha256 :no_check
+
+  url 'https://github.com/github/octicons/archive/master.zip'
+  homepage 'https://octicons.github.com'
+  license :ofl
+
+  font "octicons-master/octicons/octicons-local.ttf"
```
vitorgalvao: These need to be single quotes.

vitorgalvao:
Could you please squash your commits?

benbalter:
> Could you please squash your commits?

Done.

PR #398, opened 2015-10-13
Add font-keep-calm

rstacruz:
It's a commercial font with one of the weights available for free. http://www.1001fonts.com/keep-calm-font.html

### DIFF
--- a/Casks/font-keep-calm.rb
+++ b/Casks/font-keep-calm.rb
@@ -1,0 +1,10 @@
+cask :v1 => 'font-keep-calm' do
+  version :latest
+  sha256 '11a6079a98b53a67cd683d2b089014d385a3ea04a888d9d27408e73578a383d7'
+
+  url 'http://dl.1001fonts.com/keep-calm.zip'
+  homepage 'http://www.1001fonts.com/keep-calm-font.html'
+  license :commercial # free for personal use
+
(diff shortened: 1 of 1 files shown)


Casks/font-keep-calm.rb
```diff
@@ -0,0 +1,10 @@
+cask :v1 => 'font-keep-calm' do
+  version :latest
+  sha256 '11a6079a98b53a67cd683d2b089014d385a3ea04a888d9d27408e73578a383d7'
+
+  url 'http://dl.1001fonts.com/keep-calm.zip'
+  homepage 'http://www.1001fonts.com/keep-calm-font.html'
+  license :commercial # free for personal use
```
vitorgalvao: It’s `:freemium`, then. Please remove the comment.

Casks/font-keep-calm.rb
```diff
@@ -0,0 +1,10 @@
+cask :v1 => 'font-keep-calm' do
+  version :latest
```
vitorgalvao: Seeing as `url` doesn’t point to a specific version, this cask should use `version :latest` and `sha256 :no_check`. [See an example](https://github.com/caskroom/homebrew-cask/blob/master/Casks/acorn.rb).

PR #407, opened 2015-11-02
update sha256 for font-fontawesome

glifchits:
Sorry about that, I provided a shasum for an alternate download of the file. This one should work better.


victorpopkov:
Thank you for the fix.

PR #412, opened 2015-11-23
updated d2coding font to v1.1

herblover:
font file download link changed.
fixed sha256 hash.


adityadalal924:
Please edit your commit message to "updated d2coding font to v1.1"

http://stackoverflow.com/questions/179123/edit-an-incorrect-commit-message-in-git

herblover:
I've updated commit message. thank you!

PR #423, opened 2015-12-21
N-Gage Font

eabay:
Add N-Gage font from [dafont](http://www.dafont.com/n-gage.font)


adityadalal924:
homepage seems defunct?

eabay:
It's the author address on dafont page. I couldn't accesss too but added
for reference.

Issue #437, opened 2016-01-06
Fonts missing 'name'


vitorgalvao:
All these fonts are missing the (mandatory) `name` stanza. They should be outright deleted from this list as they are fixed. List should be rebuilt periodically (to make sure everything is up to date).

To rebuild the list (it assumes you’re inside `homebrew-fonts/Casks`):

```bash
for font in $(grep --files-without-match --extended-regexp "\s+name '.*'" *); do
  echo "+ [${font}](../blob/master/Casks/${font})"
done | pbcopy
```

+ [font-bm-hanna.rb](../blob/master/Casks/font-bm-hanna.rb)
+ [font-clara.rb](../blob/master/Casks/font-clara.rb)
+ [font-lao-muang-don.rb](../blob/master/Casks/font-lao-muang-don.rb)
+ [font-lao-muang-khong.rb](../blob/master/Casks/font-lao-muang-khong.rb)
+ [font-lao-sans-pro.rb](../blob/master/Casks/font-lao-sans-pro.rb)
+ [font-lateef.rb](../blob/master/Casks/font-lateef.rb)
+ [font-miama.rb](../blob/master/Casks/font-miama.rb)
+ [font-monoone.rb](../blob/master/Casks/font-monoone.rb)
+ [font-nanum-brush-script.rb](../blob/master/Casks/font-nanum-brush-script.rb)
+ [font-nanum-pen.rb](../blob/master/Casks/font-nanum-pen.rb)
+ [font-souliyo-unicode.rb](../blob/master/Casks/font-souliyo-unicode.rb)
+ [font-tharlon.rb](../blob/master/Casks/font-tharlon.rb)

caarlos0:
@vitorgalvao PR #439 might fix this issue...

vitorgalvao:
Done. All remaining fonts are from https://github.com/caskroom/homebrew-fonts/issues/445.

PR #439, opened 2016-01-09
Adds the name stanza for all fonts

caarlos0:
Currently, some fonts won't install, as reported in #437

I made a little shell script based on the script used to report the
problematic fonts to fix them.

It is probably not the best solution, but at least users will be able to
install the fonts.

My script is:

```sh
for font in $(grep --files-without-match --extended-regexp "\s+name'.*'" *); do
  name="${${${font#font-}%.rb}/-/ }"
  echo "$font -> $name"
  sed -i '' -e $'1 s/do/do\\\n  name '"'$name'"'/' "$font"
done
```


vitorgalvao:
> but at least users will be able to install the fonts.

Not having `name`s doesn’t stop users from installing these.

Thank you for the effort, but there is a reason why we haven’t gone with this solution. `name` isn’t just a repeat of the token, it [has a purpose and structure](https://github.com/caskroom/homebrew-cask/blob/master/doc/CASK_LANGUAGE_REFERENCE.md#name-stanza-details). We can’t automate it, we need to do it manually.

Also, [it’s in the wrong position](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#stanza-order).

caarlos0:
@vitorgalvao one example: 

```
brew cask install font-hack
Error: Cask 'font-hack' definition is invalid: Bad header line: '{:v1=>"font-hack"}' does not match file name
```

Yeah, I figured it was in wrong order, already fixed that:

```sh
for font in $(grep --files-without-match --extended-regexp "\s+name '.*'" *); do
  name="$(echo $font | sed -e 's/font-//' -e 's/-/ /g' -e 's/.rb//')"
  echo "$font -> $name"
  gsed -i'' -e '/url /a \ \ name '"'$name'" $font
done
```

but, ok...

vitorgalvao:
That error is unrelated. It's caused by an update to how `homebrew-cask` works. Details can be found [here](https://github.com/caskroom/homebrew-cask#important-december-update-homebrew-cask-will-now-be-kept-up-to-date-together-with-homebrew-see-15381-for-details-if-you-havent-yet-run-brew-uninstall---force-brew-cask-brew-update-to-switch-to-the-new-system).

To fix the issue, please run:
```bash
brew uninstall --force brew-cask; brew update
```

If you are still having problems, [open an issue with the appropriate template](https://github.com/caskroom/homebrew-cask#questions-wanna-chat)

caarlos0:
Ok, thanks.

tsparber:
Thinking about it: How should a good (font) name be formed?
For example [`font-amaranth.rb`](https://github.com/caskroom/homebrew-fonts/blob/master/Casks/font-amaranth.rb):

```
  name 'Amaranth'
  name 'Google Amaranth'
  name 'Gesine Todt Amaranth'
  name 'Google Gesine Todt Amaranth'
```
(The name comes from [Google Fonts Amaranth](https://www.google.com/fonts/specimen/Amaranth))

vitorgalvao:
In this case, `Google` is a provider, not a creator, so it must be kept out (you wouldn’t append `softpedia` or `macupdate` to an app simply because they distribute it).

Your first and third suggestions are both good depending on context. For a font that is only distributed via Google Fonts (it is essentially vendorless), the first is better, while the third is best for other cases. Always keep in mind what matters here is the vendor of the font, not the designer. So even if Erik Spiekermann designed [Fira](https://github.com/mozilla/Fira), it should still be `Mozilla Fira`.

PR #444, opened 2016-01-15
remove all fonts hosted on google fonts github repo and google code

vitorgalvao:
Repo [was there yesterday](https://web.archive.org/web/*/https://github.com/google/fonts), gone today.

This also means removing subversion support is now again a possibility (only 12 fonts remain, and they should be small repos).

Calling on @caskroom/maintainers to see if anyone has any idea why Google deleted the repo. We should still wait a few days, to make sure it isn’t just a mistake that’ll be corrected.

This should also be merged via github’s web interface, so we get a nice way to revert.


phinze:
Oh no! These were so useful. :frowning:

vitorgalvao:
> Oh no! These were so useful. :frowning: 

And it’ll get about five times worse with my next commit, since I’ll also be removing fonts that are hosted on google code, since that’ll be closed in 10 days.

victorpopkov:
Well, this is unfortunate. First Google Code, now Fonts. Didn't find any information why they have decided to remove the Fonts repository from GitHub so suddenly. :disappointed:

vitorgalvao:
Wouldn’t be surprised if this repo itself, along with other systems that used their repo for easy install, would be part of the problem:

+ They never intended for anyone to download those fonts automatically on demand. The only way they condoned getting fonts from the repo was by downloading the whole master archive. `caskroom/fonts` was exploiting a github feature of supporting svn to get them.
+ They get valuable usage data when you link google fonts from the provided methods they push on you (not manually downloading and hosting fonts).

adityadalal924:
The honebrew-fonts repo is about to get ludicrously smaller - Did Google give alternatives for downloading a single font?

jawshooah:
Shit, this tanks a pretty huge proportion of available fonts. We should definitely hold off on this for a few days in case the repo resurfaces or alternative sources are provided.

vitorgalvao:
> Did Google give alternatives for downloading a single font?

Same ones they already had. You need to pick them one by one, tick all the weights (also one by one), and then download a zip file, which will have a URL like https://www.google.com/fonts/download?kit=3hvsV99qyKCBS55e5pvb3ltkqrIMaAZWyLYEoB48lSQ, so we can’t actually automate adding them, and I’m not even sure it’ll work for long.

image

Oh, and you’ll need to do that every time there’s a change to the font.

There’s also [SkyFonts](http://www.fonts.com/web-fonts/google), which is awful.

And they still link to the dead github repo, as you can see on the image.

vitorgalvao:
> or alternative sources are provided

There were previously a few github repos that hosted them all (we used them here, even). They were left unmaintained and deleted, since Google was doing it themselves.

jimthedev:
I don't see anything about it in the [dmca](https://github.com/github/dmca) repo either. :(

davelab6:
It was a mistake on my part; I'll put it back shortly.

vitorgalvao:
@jimthedev Why would this be a DMCA issue? Google has all the rights to host those.

@davelab6 Do you mean the whole github.com/google/fonts repo will be put back up?

I’ll also point out that even with github.com/google/fonts back up, this repo will still lose almost half its fonts, due to the imminent google code shutdown.

davelab6:
> @davelab6 Do you mean the whole github.com/google/fonts repo will be put back up?

Yes. I was attempting to delete github.com/davelab6/fonts :( I've emailed support@github.com about it; if not I can create a new repo and push my local branch; I'll need to recreate issues based on the emailed copies of messages.

> I’ll also point out that even with github.com/google/fonts back up, this repo will still lose almost half its fonts, due to the imminent google code shutdown.

I don't understand this. The google code hg repo contents were moved to git in full?

vitorgalvao:
> I was attempting to delete github.com/davelab6/fonts :( I've emailed support@github.com about it;

Might also be a good time to point out to them to [Never Use a Warning When you Mean Undo](http://alistapart.com/article/neveruseawarning). Every time I delete a repo, I dread how bad their system is.

> The google code hg repo contents were moved to git in full?

For some reason, I was under the impression we had already completed the transition on all fonts, and all that remained were from other sources, but it appears that might be the case just for `cwtex`.

I’ll go ahead and close this, since the wheels of fixing are already in motion.

How did you find out about this specific issue, though? If someone contacted you and pointed you here, thank you to them.

davelab6:
I agree, but I wish I had simply been forced to confirm with `google/fonts` instead of `fonts`. 

Are any cwTex fonts missing from www.google.com/fonts/earlyaccess or https://github.com/davelab6/fonts/tree/master/ofl ?

> How did you find out about this specific issue, though?

I checked twitter for outrage at my stupidity :dancer:

vitorgalvao:
> Are any cwTex fonts missing from www.google.com/fonts/earlyaccess or https://github.com/davelab6/fonts/tree/master/ofl ?

`cwtex-q` fonts, but already found [the repo](https://github.com/l10n-tw/cwtex-q-fonts).

> I checked twitter for outrage at my stupidity :dancer: 

Hope it isn’t too bad. Best of luck with recovering. Hopefully github can get the issues back as well.

jimthedev:
@vitorgalvao It is pretty common for DMCAs to come through for fonts. It is also common for DMCAs to be sent out en-mass referring to incorrect repos. I was just double checking to make sure it hadn't happened. I am not saying that Google somehow does not have the rights, rather, was just double checking Github's work.

vyp:
If it helps, @Calinou did this: <https://github.com/Calinou/google-fonts>. Although of course it doesn't have all the issues.

davelab6:
The google project is back :)

PR #452, opened 2016-01-28
Rename libertine→libertinus per upstream and use versioned release

alerque:


Casks/font-libertinus.rb
```diff
@@ -0,0 +1,26 @@
+cask 'font-libertinus' do
+  version '6.1'
+  sha256 '2c5a32e08d7601ffdc86e52d69dc1d19acfd118ec5bb549073a3cbeb969e74be'
+
+  url "https://github.com/khaledhosny/libertinus/archive/v#{version}.tar.gz"
+  appcast 'https://github.com/khaledhosny/libertinus/releases.atom',
+          sha256: 'bf75faf5d43f2bb6324d38386cd14046fc372afc9ac8c27f3475f40c344f900f'
```
vitorgalvao: This is different, now ([docs](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md))

alerque: Dude. Every time I contribute to this repo something has changed that isn't implemented in any other items yet. I'm just copying what's been done before, if zero of the current formula use a syntax...

vitorgalvao: Well, we are pre-alpha, so things are bound to change. Thank you for noticing, though. I was under the impression we had fixed all repos, but looks like this one slipped by.

vitorgalvao: Has a heads-up, `license` will eventually be removed.

PR #453, opened 2016-01-28
Fix packaging for tex-gyre fonts that changed upstream

alerque:


vitorgalvao:
The ones that lost `#{version}` should now use `version :latest` and `sha256 :no_check`.

alerque:
@vitorgalvao None of these lost `#{version}` as that is still part of the download URL. It is just no longer part of the path that the font files extract to.

vitorgalvao:
Oh, you’re right. I was looking at your changes to `homepage`, by mistake.

Merging.

PR #454, opened 2016-01-28
Update appcast syntax per current docs

alerque:
Per [comment left on my PR](https://github.com/caskroom/homebrew-fonts/pull/452#discussion_r51116768), this updates the syntax for all the existing formulas that use appcast.

### DIFF
--- a/Casks/font-amiri.rb
+++ b/Casks/font-amiri.rb
@@ -5,7 +5,7 @@ cask 'font-amiri' do
   # github.com is the official download host per the vendor homepage
   url 'https://github.com/khaledhosny/amiri-font/releases/download/0.107/amiri-0.107.zip'
   appcast 'https://github.com/khaledhosny/amiri-font/releases.atom',
-          sha256: 'b04d8b1e4ee31bdda5987d1c6f7f4bba332a27b971966d455da32d9cc1f20dfb'
+          checkpoint: 'b04d8b1e4ee31bdda5987d1c6f7f4bba332a27b971966d455da32d9cc1f20dfb'
   homepage 'http://www.amirifont.org/'
   license :ofl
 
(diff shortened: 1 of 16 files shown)


vitorgalvao:
Thank you for this! Was about to start making the change myself.

alerque:
@vitorgalvao Start? It's only one command, I'm not sure what the middle and end points would have been.

    sed -i -e '/appcast/{N;s/sha256/checkpoint/;}' Casks/*.rb

vitorgalvao:
> It's only one command

In this case it is, but there’s no way you could know that without actually inspecting the casks. Seems like you haven’t done so, which means this change was pure luck.

`:checkpoint` isn’t just a renamed `:sha256` (there wouldn’t be much point to that), and uses a different mechanism (again, [docs](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md)). The only reason these are the same in this case is github’s automatic appcasts don’t have a `<pubDate>` tag, otherwise they would indeed be different, and need to be recalculated.

In addition, since the only thing you did was change them, it means you haven’t actually checked if those are up-to-date or not (which previously was not necessary, but now it is), which means I’ll have to actually manually check all of those anyway.

alerque:
I did inspect the casks to verify what action to take and whether it would work across oll of them. I also checked the cheksums on some of them, but since I don't know how to execute whatever code would normally be running on these it's hard to actually test. What process actually consumes the appdata information?

vitorgalvao:
> but since I don't know how to execute whatever code would normally be running on these

It isn’t relevant (and it’s still being worked on). All that is needed it that it is calculated by the one liner given in the documentation. How it will be consumed makes no difference to include it in the cask, just that it is correct.

PR #463, opened 2016-02-02
added font-latin-modern & updated font_casker header

sbdchd:


Casks/font-latin-modern.rb
```diff
@@ -0,0 +1,81 @@
+cask 'font-latin-modern' do
+  version '2.004'
+  sha256 '5b0236051d3728be6616f1b274e3b910473875b5a3ef9382f0ef00384ddb161b'
+
+  url 'http://www.gust.org.pl/projects/e-foundry/latin-modern/download/lm2.004otf.zip'
```
adidalal: use "" and `#{version}` instead of hardcoding

vitorgalvao:
In the future, please submit separate pull requests for each cask, as more often than not this type of pull request brings problems with one or more of them, and it halts the inclusion of the other ones.

At the very least these should be two commits, not one.

PR #464, opened 2016-02-02
Added latin modern math

sbdchd:


Casks/font-latin-modern-math.rb
```diff
@@ -0,0 +1,11 @@
+cask 'font-latin-modern-math' do
```
vitorgalvao: Should be `latinmodern-math` instead.

vitorgalvao: Nevermind, I see the consistency with the other font.

PR #467, opened 2016-02-04
updated font-lobster to use new location

sbdchd:


adidalal:
@vitorgalvao googlecode archive vs github/google/fonts preference?

vitorgalvao:
Preference over github, no doubt.

vitorgalvao:
Preference over github, no doubt. Feel free to merge if you agree.

adidalal:
Was asking because this *only* works because it's a single font file. Other google fonts still need to be pulled from googlecode. Either way, I figure we should be consistent

vitorgalvao:
> Other google fonts still need to be pulled from googlecode.

No need. Those are the ones that use svn, (e.g. [`abeezee`](https://github.com/caskroom/homebrew-fonts/blob/f8d649d7c706880e346a263ce996bb47e7cfe284/Casks/font-abeezee.rb)).

PR #514, opened 2016-03-01
update font-dejavu-sans to 2.35

MitMaro:
Version 2.34 was failing the checksum and the url seemed to be invalid. Bumped the version to 2.35 and updated the url.

### DIFF
--- a/Casks/font-dejavu-sans.rb
+++ b/Casks/font-dejavu-sans.rb
@@ -1,8 +1,9 @@
 cask 'font-dejavu-sans' do
-  version '2.34'
-  sha256 'f6b806b4aa0ea36a8d0f2c46dcd5afa6743773930024e562980c4cd0f9230792'
+  version '2.35'
+  sha256 '7e0d00f20080784c3a38a845d5858c161af14f0073d9474cdbfdedae883cc747'
 
   url "http://downloads.sourceforge.net/sourceforge/dejavu/dejavu-fonts-ttf-#{version}.zip"
+  name 'DejaVu'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
The old url is still valid, and is actually preferred.

Casks/font-dejavu-sans.rb
```diff
@@ -1,8 +1,8 @@
 cask 'font-dejavu-sans' do
-  version '2.34'
-  sha256 'f6b806b4aa0ea36a8d0f2c46dcd5afa6743773930024e562980c4cd0f9230792'
+  version '2.35'
+  sha256 '7e0d00f20080784c3a38a845d5858c161af14f0073d9474cdbfdedae883cc747'
 
-  url "http://downloads.sourceforge.net/sourceforge/dejavu/dejavu-fonts-ttf-#{version}.zip"
+  url "http://iweb.dl.sourceforge.net/project/dejavu/dejavu/#{version}/dejavu-fonts-ttf-#{version}.zip"
```
vitorgalvao: Trouble here is that you’re picking a specific server instead of letting sourceforge pick the best one, and that is a problem. Something better needs to be set.

vitorgalvao: There needs to be a `name `DejaVu`' line, here, or the build won’t pass.

MitMaro:
I reverted the url and just bumped the version number, the hash and added the name.

I think I hit a bad mirror when I tried the existing url. That or I did something wrong.

PR #515, opened 2016-03-03
Add TeX Gyre Pagella Math font

apuignav:
Add TeX Gyre Pagella Math font.

### DIFF
--- a/Casks/font-texgyrepagella-math.rb
+++ b/Casks/font-texgyrepagella-math.rb
@@ -1,0 +1,10 @@
+cask 'font-texgyrepagella-math' do
+  version '1.632'
+  sha256 '4c32b3094d14ccdb649c1a668871a43ee17b3fdc34192c9f0cb8792d68148f4c'
+
+  url "http://www.gust.org.pl/projects/e-foundry/tg-math/download/texgyrepagella-math-1632.zip"
+  homepage 'http://www.gust.org.pl/projects/e-foundry/tg-math'
+  license :lppl
+
+  font 'texgyrepagella-math-1632/otf/texgyrepagella-math.otf'
+end


Casks/font-tex-gyre-pagella-math.rb
```diff
@@ -0,0 +1,10 @@
+cask 'font-tex-gyre-pagella-math' do
```
vitorgalvao: This should be instead `cask 'texgyrepagella-math' do`. The [token reference](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md) needs to be followed, to avoid duplicates.

The file’s name also needs to be changed to `texgyrepagella-math.rb`.

apuignav:
I have applied your comments. Sorry for the confusion, I followed the `font-tex-gyre-*` fonts naming convention without checking further.

vitorgalvao:
We’ll need to fix those as well. You did well looking for a convention, it’s just that in this case the present casks are wrong.

Could you please [squash your commits](http://davidwalsh.name/squash-commits-git)?

apuignav:
Done!

Casks/font-tex-gyre-pagella-math.rb
alerque: I'm confused about this change. The token reference page you link specifically says:

> Spaces become hyphens.

What am I missing here?

alerque:
I'm confused about the naming convention change here. The token reference page you link specifically says:

> Spaces become hyphens.

What am I missing?

vitorgalvao:
@alerque What’s the confusion, then? There are no spaces in the file name.

alerque:
@vitorgalvao I guess where I differ is I don't think you're starting with the right piece of data. This isn't an app bundle and the upstream filename does not translate to a user-facing name in the way that "name on disk" does for an app bundle. The user facing name is the font name as it identifies itself. Run `fc-list | grep Gyre` and look at the names it reports. That's the name the project has given to the font and the most logical name to parse to come up with a token. Without the spaces (or hyphens) these fonts become an unintelligible mass of letters as opposed to having semantic meaning of the source, modification,  and specific family names.

Using the distribution file name does not make sense for fonts. If you go down this road all sorts of fonts are going to have to be changed from sensible names to obscure rubbish. Almost all fonts are distributed in files without spaces to accommodate systems where they don't work right, and often comply with 8 character restrictions, etc.

vitorgalvao:
@alerque I had no knowledge of `fc-list` (and I see other related) command.

I agree with your arguments, but here’s the one thing to keep in mind: **we need clear unambiguous rules about how to name casks, as avoiding duplicates is a priority**. That, however, does not mean we cannot have different rules for fonts in relation to app bundles, **but said rules do need to be documented**.

With that in mind, I invite you (if you’re interested) to draft up clear easy to apply rules for fonts that we can discuss. “Easy to apply” is a must, here, and it’d be great if that also meant “easy to verify”, for maintainers, but start with what you think is right and we can continue the discussion from that.

PR #522, opened 2016-03-07
Update cask naming spec and rename fonts to match

### LINKED ISSUE #521: Draft naming spec and rename
Per the discussion that started in #515, I think the old seat-of-the-pants naming was working out better that the letter-of-the-law names without hyphens recently kludged in from the cask naming standards.

I propose _not_ pretending font file names are the same as Mac app names "as found on disk" and using the font face name as specified by the font itself instead.

- [x] Document how font casks should be named (documentation existed in CONTRIBUTING.md, it just needed tweaking)
- [x] Rename existing font casks to match (recently renamed fonts that don't comply reverted, others to be reset on an as-spotted bases)

alerque:
Fixes #521.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,7 @@ predictable.
 ### Start From the Font's Canonical Name
 
 The canonical font name is the font family name as returned by the command
-`otfinfo --family`.
+`fc-query --format='%{family}'`.
 
 If there is more than one family in the distribution, use your judgment to
 choose the "most famous" one.  If there is more than one style, choose the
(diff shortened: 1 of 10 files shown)


vitorgalvao:
Thank you.

CONTRIBUTING.md
```diff
@@ -81,7 +81,7 @@ predictable.
 ### Start From the Font's Canonical Name
 
 The canonical font name is the font family name as returned by the command
-`otfinfo --family`.
+`fc-query -f '%{family}'`.
```
vitorgalvao: Long flags are always better for documentation and scripts. Do instead `fc-query --family '%{family}' <pat_to_font_file>`.

alerque: @vitorgalvao It's not `--family` it's `--format`, but I'll replace it with the long option.

alerque:
Edited to use long-format option syntax...

PR #523, opened 2016-03-08
Add cask font-source-code-pro-powerline.

knpwrs:


vitorgalvao:
https://github.com/caskroom/homebrew-fonts/blob/master/Casks/font-sauce-code-powerline.rb

PR #530, opened 2016-03-14
Update font-calligraffitti.rb

gmkey:


vitorgalvao:
It’s missing [`name`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/name.md) ([many fonts are](https://github.com/caskroom/homebrew-fonts/issues/437)).

PR #540, opened 2016-04-01
Fix ricty diminished

neronplex:
updated font version.
fix font license.
use version stanza.

PR #541, opened 2016-04-02
fix Ricty Diminished font

neronplex:
#540 re-submit
updated font version.
fix font license.
use version stanza.


vitorgalvao:
Thank you for fixing it.

PR #544, opened 2016-04-06
font-fantasque-sans-mono v1.7.1 (updated formula)

mdong:
Updated link and check sum


mdong:
Sorry, a patch did not commit. Let me do this again

PR #546, opened 2016-04-06
Update font-nika.rb

sijad:
It's my first commit please recheck SHA256 and checkpoint


vitorgalvao:
Everything was correct. Thank you.

PR #548, opened 2016-04-12
Update fontawesome to 4.6.0

paul-howard:
Looks like the folks over at FortAwesome pushed an update a few hours back that breaks the `font-fontawesome` cask. Hopefully this fixes it, but I'm not 100% sure I'm using the `font_casker` utility correctly, so I'd appreciate it if someone could verify the SHA256 value? Thanks!

### DIFF
--- a/Casks/font-fontawesome.rb
+++ b/Casks/font-fontawesome.rb
@@ -1,11 +1,11 @@
 cask 'font-fontawesome' do
-  version '4.5.0'
-  sha256 '6eb9dd94a87ebe10e75acb78a75530ccb33bb60a3b28d0a2c72f14fceecca18a'
+  version '4.6.0'
+  sha256 '4f9b528e02e90e24782383050ae5b69c9a73de1734796bf8763ee2d3a128c8ac'
 
   url "https://fortawesome.github.io/Font-Awesome/assets/font-awesome-#{version}.zip"
   name 'Font Awesome'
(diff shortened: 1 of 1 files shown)


sweiz:
if we update the url to `https://github.com/FortAwesome/Font-Awesome/archive/v#{version}.zip` instead of the current github pages url it won't break on a new version release in future.

vitorgalvao:
@paul-howard You can do `brew cask audit --download {{path_to_you_cask}}`. If the `sha256` is wrong, it’ll tell you.

paul-howard:
Thanks, @vitorgalvao. That's perfect.

adidalal:
@vitorgalvao Worth switching over to the URL provided by @sweiz?

vitorgalvao:
@adidalal If it does prevent the cask from breaking constantly, I’m not opposed to it. We should add an `appcast` to the cask, also.

vitorgalvao:
@adidalal On second thought, since we can add an `appcast` to the cask, I’d rather keep the official URL.

jiegec:
@victorpopkov Now the 4.6.1 is out so the link can not be used now.

sweiz:
The file at https://github.com/FortAwesome/Font-Awesome/releases.atom doesn't reference the assets url, nor from what I can see, the checksum, and my fork with the updated url appears to work with the github assets url. 

Am I missing something, or is it possible to use the `appcast` stanza _and_ the `releases` url?

PR #638, opened 2016-09-24
Update font-tex-gyre-pagella-math: bug fix

ickc:

I think the source silently updated the folder structure.

As for the audit, it has an error, which I don't understand.

```bash
audit for font-tex-gyre-pagella-math: failed
 - at least one name stanza is required
Error: audit failed for casks: font-tex-gyre-pagella-math
```

### DIFF
--- a/Casks/font-tex-gyre-pagella-math.rb
+++ b/Casks/font-tex-gyre-pagella-math.rb
@@ -1,9 +1,9 @@
 cask 'font-tex-gyre-pagella-math' do
   version '1.632'
   sha256 '68a9c0ce195915334673960b13a281f24d31a8ae380454a0e35652dba506d474'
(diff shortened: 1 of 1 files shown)


vitorgalvao:
> As for the audit, it has an error, which I don't understand.

It means, add something [like this](https://github.com/caskroom/homebrew-fonts/blob/63e937a97741c1036bec9f1b1562a96e38454d70/Casks/font-underdog.rb#L6).

ickc:
I figured out a minute ago, but obviously not as fast as you do.

vitorgalvao:
You need to rerun the commands in the template. You’ll notice `name` is in the wrong line.

ickc:
You're right. Done.

PR #648, opened 2016-10-14
Auto fix rubocop

maxnordlund:
I've fixed a bunch of warnings emitted by RuboCop, and from having

```
1044 files inspected, 798 offenses detected
```
it's down to
```
1044 files inspected, 24 offenses detected
```

The commits are unfortunately very large as a result, but each are a mechanical change. I used a `sed` script, because that's what I'm used to:

```sh
 sed -i "" '/^$/N;/\n *url .*<URL_SECTION>s/^/\
  # <URL_SECTION> was verified as official when first introduced to the cask/' Casks/*

# Yes, there's a newline because mac sed don't understand \n
```

*If there’s a checkbox you can’t complete for any reason, that's okay, just explain in detail why you weren’t able to do so.*

- [ ] `brew cask audit --download {{cask_file}}` is error-free.
- [ ] The commit message includes the cask’s name and version.

- [ ] Named the cask according to the [token reference].
- [ ] `brew cask install {{cask_file}}` worked successfully.
- [ ] `brew cask uninstall {{cask_file}}` worked successfully.
- [ ] Checked there are no [open pull requests] for the same cask.
- [ ] Checked that the cask was not already refused in [closed issues].

[token reference]: https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/caskroom/homebrew-fonts/pulls
[closed issues]: https://github.com/caskroom/homebrew-fonts/issues?q=is%3Aissue+is%3Aclosed


Casks/font-source-han-sans.rb
```diff
@@ -3,8 +3,8 @@
   sha256 'b684c9b659ec4998e72d22ff5ed3f531d31d506ae00a48ec72b9846be23e7c10'
 
   url 'https://github.com/adobe-fonts/source-han-sans/raw/5f5311e71cb628321cc0cffb51fb38d862b726aa/SuperOTC/SourceHanSans.ttc.zip'
-  homepage 'https://github.com/adobe-fonts/source-han-sans'
-  name "Source Han Sans"
+  name 'Source Han Sans'
+  name 'Source Han Sans'
```
jawshooah: Fixed in https://github.com/caskroom/rubocop-cask/pull/39. Will merge and push up a new release soon.

maxnordlund:
OK, so I've aded the name stanza to a bunch of formulae, and reduce the audit errors to 32. But I don't think this should be blocked by that as it's an improvement for sure, even though it doesn't fix everything.

Sounds OK, @jawshooah?

vitorgalvao:
Did you add `name` with a script? A bunch of them are duplicated, and others are plain wrong (`Alegreya Sans%20SC`).

We have [a method to naming these](https://github.com/caskroom/homebrew-fonts/blob/master/CONTRIBUTING.md#naming-font-casks), and it must be followed. We had other other attempts at auto-naming via a script in the past, but none was acceptable, hence why so many casks still lack `name` — it’s not dead simple.

Kind of the same reason so many casks lacked the comment. The comment says the download host was verified, so now a maintainer *has to manually verify each of those*, because even if we trust you (and we do), your script did not do any verification, it simply added the comment, which means the comment is inaccurate. Only because most of them have the same `homepage` and `url` pair do we know we can skip most (since the google ones we know to be correct).

maxnordlund:
Yes, I did use a script, and while I totally agree that it would be impossible to do this in the general case, I think it's doable in the specific case here. Maybe the naming was a bit much, and I can easily remove it (yay for clean commits). I did read that and tried to follow it, and looked at a few of the other fonts. They seem to follow a simple scheme, like [Zeyada](https://github.com/caskroom/homebrew-fonts/blob/master/Casks/font-zeyada.rb). I also tried to scope it to just he Google fonts, but I guess my script was a fair bit to overzealous.

I did verify each homepage/url pair, but obviously not each  one. Again, I focused on the Google fonts, but as you can see a did a couple more. Which all can be removed. Each script run, aka homepage/url pair, is in a separate commit, so it should be easy to verify. Well, I'd hoped so.

---

As a concrete course of action, should I just remove all non-google fonts? And/or the naming? And/or try to redo the naming and be much more careful?

vitorgalvao:
Regarding the comment — and to be absolutely clear — **your course of action was absolutely correct**. Like I said, since most of these are google pairs, it’s actually not much work to verify. And if you checked the non-google ones, so much the better.

So as for the comment, everything is good.

---

My concern was regarding only the naming. As it currently stands, it can’t be merged due to the errors (`%20` and perhaps others, as well as the duplicates). I was also concerned with correct naming, but thinking more about it, I think we can actually scale on that a for this case. Correct naming is meant to be used in `brew cask search` (not yet implemented the feature to search `name`) and disambiguate. But now that I think about it, I don’t ever recall a confusion with fonts regarding their token/name (and I have been with HBC since before this repo and the `font` stanza even existed), so if you refine your script to take care of those problems, I’d say we can even take it like that and fix later any (unlikely) mismatches users find.

maxnordlund:
> Regarding the comment — and to be absolutely clear — your course of action was absolutely correct. 

Thank you ❤️ 

---

So I redid the naming using a ruby script this time instead of sed, and manually verified the names. Some were broken, so I just removed them, but the rest should be good. They are only the google ones.

I also used the list of failed casks from audit, so there shouldn't be any duplicates this time.

```ruby
require "uri"
lines = File.open(ARGV[0]).read.split "\n"

index = lines.find_index { |line| line.include? "homepage" }
name = URI.decode(lines[index].split("/").last.chop)

lines.insert index, "  name '#{name}'"
puts lines.join "\n"
```

maxnordlund:
I've reverted/fixed the Casks/font-source-han-sans.rb as well, just realized the homepage was gone.

vitorgalvao:
Thank you @maxnordlund. Now to greatly reduce https://github.com/caskroom/homebrew-fonts/issues/437.

PR #793, opened 2016-10-20
Delete font-meslo-lg.rb

vitorgalvao:
For some reason, this one is not being able to download.


vitorgalvao:
Merged as 0c518e2c369a2dc44e74ff97d63e7b53de716742.

dkim:
@vitorgalvao Can we reinstate this formula? I have just checked that the fonts are successfully downloaded with this formula.

vitorgalvao:
@dkim [Yes](https://github.com/caskroom/homebrew-fonts/pull/950).

Also, we call these casks, not formulae. They have different syntax, purpose, and behaviour.

PR #905, opened 2016-11-01
CONTRIBUTING.md: Update cask examples

guerrero:
Now that `license` stanza has been removed and string variable interpolation now can be used, I think that both cask examples should be updated to avoid new users feel confused.

Maybe use a link to each cask file could be a great alternative but IMHO it's better to show the examples inline because user can read everything without navigate to other files


vitorgalvao:
Thank you.

Issue #909, opened 2016-11-04
Tap fails due to invalid formulas


bracki:
On a fresh installation tapping currently fails:

```
> brew tap caskroom/fonts
==> Tapping caskroom/fonts
Cloning into '/usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts'...
remote: Counting objects: 1054, done.
remote: Compressing objects: 100% (613/613), done.
remote: Total 1054 (delta 698), reused 484 (delta 438), pack-reused 0
Receiving objects: 100% (1054/1054), 191.86 KiB | 0 bytes/s, done.
Resolving deltas: 100% (698/698), done.
Checking connectivity... done.
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-cantora-one.rb
undefined method `cask' for Formulary::FormulaNamespaceb32d0debd2948a643013f407c1d4ab45:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-hammersmith-one.rb
undefined method `cask' for Formulary::FormulaNamespacebb0be1ac24f725847ab5fabdcbb70e55:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-headland-one.rb
undefined method `cask' for Formulary::FormulaNamespace0318ae2e9b307a46539a5fc7ebbc3818:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-moulpali.rb
undefined method `cask' for Formulary::FormulaNamespacee021b5f204517360e9270f5401ceb832:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-odor-mean-chey.rb
undefined method `cask' for Formulary::FormulaNamespace25ba4fe85d9c888a2a921c70829bd5bd:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-preahvihear.rb
undefined method `cask' for Formulary::FormulaNamespace09b8728c9880d77b0705fd8ceb2c4544:Module
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/caskroom/homebrew-fonts/font-sirin-stencil.rb
[...]

PR #910, opened 2016-11-04
Fix: Move previous created casks from root to /Cask folder

guerrero:
This commit fixes issue #909


vitorgalvao:
@reitermarkus Shouldn’t Travis have caught this? It used to (if I recall correctly).

PR #917, opened 2016-11-15
font-fira-sans 4.203

alanorth:

Kinda major update to Fira Sans (4.1→4.2). The cask formula changes quite a bit because some of the links to the project website are dead, and the upstream package itself now includes three font variants: Compressed, Condensed, and Normal. I only include the "Normal" fonts in this cask because that's what was there before, but I guess they could be included as well. What do you think?


vitorgalvao:
Include all the variants.

alanorth:
Latest commit includes Compressed, Condensed, and Normal variants.

vitorgalvao:
Looking further into it, this should instead use `https://mozilla.github.io/Fira/` as the homepage, have an [`appcast`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md) pointing to `https://github.com/mozilla/Fira/releases.atom`, and `url` from [the releases page](https://github.com/mozilla/Fira/releases).

Same should be done for [Fira Mono](https://github.com/caskroom/homebrew-fonts/blob/master/Casks/font-fira-mono.rb).

Mind making the changes to both casks?

alanorth:
I did notice that there was a GitHub repo for Fira, but saw that the release was one minor version behind the type foundry's release (4.202 vs 4.203). I agree that these should be using appcasts as is the new convention with Homebrew.

I will make the changes to both Fira Sans and Fira Mono. You can close this and I will submit them both in new pull requests, if you'd like.

vitorgalvao:
Thank you.

alanorth:
Actually, I don't think we should use appcasts for Fira Sans and Fira Mono. The [releases page](https://github.com/mozilla/Fira/releases) of the Fira repository on GitHub doesn't distinguish between Sans and Mono releases, so appcasts aren't particularly useful in this case. Also, the versions there are lagging behind the versions on the [foundry's website](https://carrois.com/projects/Fira/).

vitorgalvao:
> The releases page of the Fira repository on GitHub doesn't distinguish between Sans and Mono releases

That’s not a problem. We have other similar cases, and it works fine. It’s not like these are updated every week. Having the appcast is way more beneficial.

> Also, the versions there are lagging behind the versions on the foundry's website.

Not by much, and there’s no reason to believe they won’t be updated soon, now that they were updated on the foundry. Ultimately, this is Mozilla’s font, so their site and their repo should take precedence in the casks as well.

PR #921, opened 2016-11-16
Update font-fira-mono.rb

alanorth:

Updates the cask style to use the official Mozilla appcast and homepage. We cannot use `#{version}` interpolation because the Mozilla releases do not distinguish well between Sans and Mono versions, but the 4.202 tarball _does_ contain Mono 3.206.


alanorth:
`#{version.after_comma}` is genius. 👏

PR #945, opened 2017-01-04
Create font-langdon.rb

miccal:
Repeat of https://github.com/caskroom/homebrew-fonts/pull/942 added to the right directory.

### DIFF
--- a/Casks/font-langdon.rb
+++ b/Casks/font-langdon.rb
@@ -1,0 +1,10 @@
+cask 'font-langdon' do
+  version :latest
+  sha256 :no_check
+
+  url 'http://www.ffonts.net/Langdon.font.zip'
+  name 'Langdon'
+  homepage 'http://www.ffonts.net/Langdon.font'
+
+  font 'Langdon.otf'
+end


vitorgalvao:
We really need to fix that test to check for the correct directory.

miccal:
Ping @jawshooah.

jawshooah:
Unfortunately I think someone else will have to take care of it if it needs taking care of sooner rather than later. I've had no time for OSS the past couple months, and that's not likely to change for a couple more 😞 

My original attempt is [here](https://github.com/caskroom/homebrew-cask/pull/26819), but there were false positives so it was reverted [here](https://github.com/caskroom/homebrew-cask/pull/26854).

miccal:
Thanks for your reply @jawshooah - unfortunately I am quite an amateur at Ruby (I do most of my coding in Maple and Matlab) but I will open an issue and hopefully get some help.

jawshooah:
You're in luck, the relevant code is actually just Bash 😉

miccal:
Just as amateur in Bash!

vitorgalvao:
I’m pretty competent with bash, but I cannot see what the problem is. I’m looking at it at everything looks fine.

We might have to merge that again just to see if I can diagnose the problem when it actually occurs.

vitorgalvao:
It occurred to me the only thing that might be wrong there is Travi’s commit range, so I did some digging. Found [this](https://github.com/travis-ci/travis-ci/issues/4596) and [this](https://github.com/jawshooah/homebrew-cask/commit/64dd9b240e7c29e7499eca7cf157f5ae60c58822). Might be the solution.

jawshooah:
From my [comment](https://github.com/caskroom/homebrew-cask/pull/26819#issuecomment-261253755) on the original PR:

> The problem is that git diff *.rb is pulling in files outside the current working directory. Using ./*.rb instead doesn't seem to change anything.

vitorgalvao:
> The problem is that `git diff *.rb` is pulling in files outside the current working directory.

Not for me when testing locally. Might this have been something in `git` that changed since?

vitorgalvao:
Still running tests, but trying now with [`find` to limit depth](https://github.com/caskroom/homebrew-cask/blob/9a65035cbf11a22381340674a626c20f2d35fb92/ci/travis/script.sh#L14).

vitorgalvao:
Alright, [it is done](https://github.com/caskroom/homebrew-cask/pull/28635). Did a few tests in that PR (since squashed) to trigger failures and successes, and it seems to be working fine. Will merge and lets see.

PR #967, opened 2017-04-02
Create font-meslo-for-powerline.rb

miccal:
Closes https://github.com/caskroom/homebrew-fonts/issues/963 and https://github.com/caskroom/homebrew-fonts/issues/964.


vitorgalvao:
Isn’t the dotted version missing?

PR #1182, opened 2017-06-13
Update font-mononoki-nerd-font-mono to 1.0.0

giannitm:

Additionally, if **updating a cask**:

- [ ] [If the `sha256` changed but the `version` didn’t](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256),
      provide public confirmation by the developer: {{link}}

### DIFF
--- a/Casks/font-mononoki-nerd-font-mono.rb
+++ b/Casks/font-mononoki-nerd-font-mono.rb
@@ -4,7 +4,7 @@ cask 'font-mononoki-nerd-font-mono' do
 
   url "https://github.com/ryanoasis/nerd-fonts/releases/download/v#{version}/Mononoki.zip"
   appcast 'https://github.com/ryanoasis/nerd-fonts/releases.atom',
-          checkpoint: '034cdfeb4f6a152086231f8388940cd9de7d8190aecbb26ec94e7a381bf7a3b7'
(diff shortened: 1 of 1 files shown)


laughedelic:
I don't understand, what is the difference of this formula with the one from #1181??

giannitm:
@laughedelic They are two different fonts – this one is a "mono" version of the font in #1181. I believe "mono" means "monospaced", and a monospaced font is one in which all the characters occupy the same amount of space (just did a Google search and that was the explanation).

laughedelic:
OK. Thanks for reply. I understand that mono means monospaced. But the original mononoki font is already monospaced, so it was unclear for me why there are two versions of the patched font. But I found an answer in the [nerd-fonts FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting):

> If you are limited to monospaced fonts (because of your terminal, etc) then pick a font with the 'Mono' suffix.
> - This denotes that the Nerd Font glyphs will be monospaced not necessarily that the entire font will be monospaced

You can see here that any letters are monospaced in both variants, but the icons are different: the first one is Mono, so its icons match the letters width:

screen shot 2017-08-04 at 3 56 56 pm

PR #1373, opened 2017-11-14
Update font-alegreya & font-alegreya-sans to upstream

cdlm:
Caveat: both now include the small caps version, making `font-alegreya-sc` and `font-alegreya-sans-sc` obsolete. I've added `conflicts_with` stanzas, but maybe it's better to just remove those two casks?

Also please tell me if I should split this into two separate pull-requests.


commitay:
> I've added `conflicts_with` stanzas, but maybe it's better to just remove those two casks?

Yes, please remove the 2 `sc` Casks and remove the conflict.

vitorgalvao:
Fine as it is. Thank you.

PR #1620, opened 2018-03-19
Revert nerd fonts to 1.2.0

commitay:
`2.0.0` zips have been removed and the tag is now pre-release.


Thammachart:
@commitay Thank for the quick fix. It was too late at my place. Too sleepy for me.
Anyway, do you have any script to automate all of these in one pull request. Formerly, I used chain of `cask-repair`. it created a lot of pull requests, and It seems not proper for me.

ryanoasis:
Sorry about this. Thanks for the fix. :medal_sports: 

I thought just removing the archives would be best when large flaws were uncovered.

PR #1853, opened 2019-07-09
Analytics cask purge

vitorgalvao:


mscottford:
If anyone is looking for the Rambla font and happens to stumble across this pull request, I created another font cask repo which has the rambla cask in it. It's a copy and paste of the one that was deleted here. https://github.com/corgibytes/homebrew-cask-fonts

vitorgalvao:
@mscottford Or, if you need the font, make a pull request to readd it, at which point it’ll be exempt from the system for another year.

I’m also thinking it might be a good idea to exclude this repo from the system, but I’d like to do a few more passes first.

mscottford:
Thanks for the tip. I just figured if it got removed for low analytics, then perhaps there weren't enough people using it, and if I wanted it to live longer then I'd need to create a new tap for that. I'll submit the pull request.

Also, when you say "exclude this repo from the system" are you talking about the font casks repo being excluded from the system that removes unused casks? If so, I would suggest either removing it or at least reducing the thresholds. Fonts don't get installed that often in general, so I'm guessing that the metrics for all of the font casks are pretty low relative to most other kinds of binaries. Just my 2 cents.

And thanks again for the response and for helping to maintain this project.

vitorgalvao:
> Also, when you say "exclude this repo from the system" are you talking about the font casks repo being excluded from the system that removes unused casks?

Yes.

> Fonts don't get installed that often in general, so I'm guessing that the metrics for all of the font casks are pretty low relative to most other kinds of binaries.

I agree.

> And thanks again for the response and for helping to maintain this project.

Thank you for the kind words.

PR #1921, opened 2019-12-10
Delete font-inconsolata-dz.rb

vitorgalvao:
Homepage is dead.

```
font-inconsolata-dz (added 2113 days ago)
30 days: 17 (#1811)
90 days: 101 (#1306)
365 days: 384 (#1259)
```


Mattsi-Jansky:
Bit disappointed as I was using this. Also, `font-inconsolata-dz-for-powerline` still exists; should probably delete both for consistency?

vitorgalvao:
> Also, `font-inconsolata-dz-for-powerline` still exists; should probably delete both for consistency?

`font-inconsolata-dz-for-powerline` is an order of magnitude more popular and its homepage isn’t down. It shouldn’t be removed, no.

PR #1933, opened 2020-01-15
Add JetBrains Mono

RiKap:
- [x] Named the cask according to the [token reference].
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.
- [x] Checked there are no [open pull requests] for the same cask.
- [x] Checked the cask was not [already refused].
- [x] Checked the cask is submitted to [the correct repo].

[token reference]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md
[open pull requests]: https://github.com/Homebrew/homebrew-cask-fonts/pulls
[already refused]: https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=Issues
[the correct repo]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask
[version-checksum]: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256

--------------

I asked JetBrains if they can provide direct link to lates version. Something like: `https://download.jetbrains.com/fonts/JetBrainsMono-lates.zip`

### DIFF
--- a/Casks/font-jetbrains-mono.rb
+++ b/Casks/font-jetbrains-mono.rb
@@ -1,0 +1,18 @@
+cask 'font-jetbrains-mono' do
+  version '1.0.0'
+  sha256 '82bf0dec956e4ca9afa4da4978fa5de80a75a576b3353d570ccbaabe4e858389'
+
+  url "https://download.jetbrains.com/fonts/JetBrainsMono-#{version}.zip"
+  appcast 'https://www.jetbrains.com/lp/mono'
+  name 'JetBrains Mono'
+  homepage 'https://www.jetbrains.com/lp/mono'
+
+  font 'JetBrainsMono-Bold-Italic.ttf'
+  font 'JetBrainsMono-Bold.ttf'
+  font 'JetBrainsMono-ExtraBold-Italic.ttf'
+  font 'JetBrainsMono-ExtraBold.ttf'
+  font 'JetBrainsMono-Italic.ttf'
+  font 'JetBrainsMono-Medium-Italic.ttf'
+  font 'JetBrainsMono-Medium.ttf'
+  font 'JetBrainsMono-Regular.ttf'
+end


Casks/font-jetbrains-mono.rb
```diff
@@ -0,0 +1,17 @@
+cask 'font-jetbrains-mono' do
+  version '1.0.0'
+  sha256 :no_check
```
rock3r: For `1.0.0` the `sha256` value should be `82bf0dec956e4ca9afa4da4978fa5de80a75a576b3353d570ccbaabe4e858389` — this is a versioned URL so it's ok to use the SHA here I believe

vitorgalvao: >  this is a versioned URL so it's ok to use the SHA here I believe

Not only OK, but necessary.

vitorgalvao:
> I asked JetBrains if they can provide direct link to lates version.

Our users prefer versioned URLs.

Casks/font-jetbrains-mono.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  url "https://download.jetbrains.com/fonts/JetBrainsMono-#{version}.zip"
```
vitorgalvao: Please add an [`appcast`](https://github.com/homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md).

Also, when submitting PRs, leave on the option for maintainers to edit directly. It considerably speeds up the process and wastes less time.

rock3r: I have asked if we can put the releases on the GitHub repo releases, so we can get the feed out of it for the `appcast`. As far as I know there's no release feed right now.

vitorgalvao: Forgot to mention it (usually I’d just do it but I can’t edit this PR directly), but the homepage can be used as the appcast for now.

RiKap:
@vibraphone Updated based on iformations mentioned here

vitorgalvao:
Thank you, @RiKap.

@rock3r If you can still get it on Github with releases, or any structured appcast (i.e. not an HTML page), we’ll still take it.

rock3r:
@vitorgalvao looks like they started using GitHub releases, so you can use them for appcasts: https://github.com/JetBrains/JetBrainsMono/releases

Issue #1961, opened 2020-02-17
font-eeyek-unicode download location has been moved


forivall:
#### General troubleshooting steps

- [x] I understand that [if I ignore these instructions, my issue may be closed without review](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/closing_issues_without_review.md).
- [ ] I have retried my command with `--force`.
- [ ] I ran `brew update-reset && brew update` and retried my command.
- [ ] I ran `brew doctor`, fixed as many issues as possible and retried my command.
- [x] I have checked the instructions for [reporting bugs](https://github.com/Homebrew/homebrew-cask#reporting-bugs).
- [x] I made doubly sure this is not a [checksum does not match](https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md) error.

#### Description of issue

`brew cask install font-eeyek-unicode` results in:
```
==> Downloading http://tabish.freeshell.org/eeyek/eeyek.zip

curl: (7) Failed to connect to tabish.freeshell.org port 80: Connection refused
Error: Download failed on Cask 'font-eeyek-unicode' with message: Download failed: http://tabish.freeshell.org/eeyek/eeyek.zip
```

Looks like it's been moved to http://tabish.freeshell.org/eeyek/eeyek-ofl.zip (see the [download page](http://tabish.freeshell.org/eeyek/download.html)). Discovered via https://github.com/Homebrew/brew/pull/7043 .

As i'm going through and installing _all the fonts_, i've also discovered the following
`brew cask install font-accordance`
```
==> Downloading https://www.accordancefiles1.com/products/Accordance_Font.zip

curl: (22) The requested URL returned error: 404
Error: Download failed on Cask 'font-accordance' with message: Download failed: https://www.accordancefiles1.com/products/Accordance_Font.zip
```
Its [download page](https://www.accordancebible.com/Accordance-Font/) results in a 404, so that may be temporary.

`font-asap`
```
==> Downloading https://www.omnibus-type.com/wp-content/uploads/Asap.zip
######################################################################## 100.0%
==> No SHA-256 checksum defined for Cask 'font-asap', skipping verification.
==> Installing Cask font-asap
==> Purging files for version latest of Cask font-asap
Error: It seems the Font source '/usr/local/Caskroom/font-asap/latest/Asap/Asap-SemiBoldItalic.ttf' is not there.
```

`font-grenze`
```
[...]

PR #1972, opened 2020-03-09
Delete font-asap.rb

vitorgalvao:
[Broken an unpopular](https://github.com/Homebrew/homebrew-cask-fonts/issues/1961).


apropostech:
Right. So one person didn't have time to fix it and it's deleted - with those with existing installs having to [manually delete](https://github.com/Homebrew/homebrew-cask/issues/58352) the installed package to get brew cask to work again.

This hardly seems to be the right approach. If it would at least be left, someone will probably fix it in due course and more importantly it wouldn't break brew cask for existing users.

vitorgalvao:
> So one person didn't have time to fix it and it's deleted - with those with existing installs having to [manually delete](https://github.com/Homebrew/homebrew-cask/issues/58352) the installed package to get brew cask to work again.

That only happens in extremely old installs—several years old.

> This hardly seems to be the right approach.

So your suggestion is to leave a cask broken indefinitely? Why isn’t your approach to readd the cask, *fixed*? We’re receptive to that.

> If it would at least be left, someone will probably fix it in due course

No, they probably wouldn’t. I’ve been maintaining Homebrew Cask since it started and I’ve experienced first-hand the abandonment of unpopular casks.

That attitude of “someone will probably fix it” is the problem, because it’s implied to mean “someone other than me”. *Nobody cares for this cask; we know it was unpopular.* There’s no “someone else”—it had to be you, the one person who cared. And if you don’t care enough to do it, please also don’t impose your will on the time of volunteers.

> and more importantly it wouldn't break brew cask for existing users.

If your cask was so old that it broke with that error, it only goes to show how untainted it was. Who knows for how long it was broken; clearly not enough people care for it.

We remove bad casks all the time and you haven’t said a thing—presumably because you were not affected—*because the vast majority aren’t*. Removing casks is necessary to keep a healthy project. If we followed your suggestion, the project would be ubmaintainable by now.

When commenting on an open-source project, I urge you to not think of just yourself. Especially if you’ve never interacted or contributed, you will lack necessary information.

apropostech:
I'm sorry if my comment came across as selfish.

I'm a regular contributor updating casks and would have updated this too if it would have been clear to me that there was an issue - but it's not "my cask". I didn't see your discussion until I noticed the error due to it being deleted. Fonts are not often updated so doesn't seem so strange that it's an old cask.

You are right of course that I don't know all the details of what's going on. Just seemed like an odd way of handling it with an apparent expectation that these kind of issues will be spotted by everyone on github.

As a suggestion, maybe there could be a message in terminal to flag a need for update when it's the case for certain casks that are installed on the system. I'm sure that would give a much better response with people noticing that there is an issue and updating it.

vitorgalvao:
> I'm sorry if my comment came across as selfish.

Accepted. Thank you.

> would have updated this too if it would have been clear to me that there was an issue

I’ve thought about some kind on newsletter that informs of casks with problems so users that use them can offer fixes, but I’m not sure how to best go about it.

> but it's not "my cask".

I didn’t say it was yours, I said you were the only one who cared about it.

> with an apparent expectation that these kind of issues will be spotted by everyone on github.

Quite the contrary. I didn’t expect that everyone would notice, but that almost no one would (which is still the case). We remove casks all the time, and only a tiny fraction get complained about *because they are unpopular*.

> As a suggestion, maybe there could be a message in terminal to flag a need for update when it's the case for certain casks that are installed on the system. I'm sure that would give a much better response with people noticing that there is an issue and updating it.

That would inundate the terminal with messages most users don’t care for and would thus ignore. By their very nature, unpopular casks are the ones nobody fixes.

apropostech:
Perhaps it was unclear, but my proposal was to only show terminal messages for those users that have installed casks that needs fixing. Presumably some of those people that have installed "unpopular casks" do actually care about them and would fix them if they knew there's an issue.

vitorgalvao:
That makes sense, but it would take an inordinate amount of work to add the deprecation, then remove after X time, and whatnot.

Let’s leave that idea in the backburner. https://github.com/Homebrew/brew/issues/7139 is tangential, so let’s see how that works first after it’s implemented and rethink your system then.

PR #2077, opened 2020-06-04
Add font-elstob 1.003

adunning:

- [X] `brew cask audit --download {{cask_file}}` is error-free.
- [X] `brew cask style --fix {{cask_file}}` reports no offenses.
- [X] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.
- [X] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

- [X] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [X] `brew cask install {{cask_file}}` worked successfully.
- [X] `brew cask uninstall {{cask_file}}` worked successfully.
- [X] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=Issues).
- [X] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).

### DIFF
--- a/Casks/font-elstob.rb
+++ b/Casks/font-elstob.rb
@@ -1,0 +1,12 @@
+cask 'font-elstob' do
+  version '1.003'
+  sha256 '52d4f5375badf6f412c84d55b1ab57691897b4d1084a140995169e4f62ac10a4'
+
+  url "https://github.com/psb1558/Elstob-font/releases/download/v#{version}/Elstob_v#{version.dots_to_underscores}_web.zip"
+  appcast 'https://github.com/psb1558/Elstob-font/releases.atom'
+  name 'Elstob'
+  homepage 'https://github.com/psb1558/Elstob-font'
+
+  font 'Elstob_font_web/variable/Elstob-Italic[GRAD,opsz,wght].ttf'
+  font 'Elstob_font_web/variable/Elstob[GRAD,opsz,wght].ttf'
+end


adunning:
The downloaded archive provides variable-width fonts, OTF, and TTF. This installs only the first, with a separate version available in #2078 to provide compatibility with applications that do not support these. Let me know if you think there's a better way to go about this.

vitorgalvao:
We should be installing the `desktop` fonts, not the `web` fonts, so going with those.

Thank you for being proactive and doing both PRs.

adunning:
It's a slightly weird approach to packaging. Along with versions intended for 'web' use, the package also includes variable fonts (what I have included here) that seem to be identical to the desktop fonts, and function more cleanly within macOS applications that use system font support. The 'desktop' versions are only necessary for LaTeX and some older applications that do not support OpenType variable fonts. I've therefore installed both, but would be interested to know if I am misunderstanding.

vitorgalvao:
I’m unsure here. I haven’t been keeping up with variable font tech, so I’m not sure how ready or not they are. Might be worth opening an issue in the main repo to discuss which to include.

We’re not going to provide both, the question is “should we start defaulting to variable fonts, or is it too soon?”

adunning:
I've done more research into this with #2101 – given that there are several other fonts doing something similar, I feel it would be useful to include both Elstob and ElstobD (on the model of the Source families and others). Alternatively, if you think such duplication is unacceptable, I think that only this variable version should be provided.

vitorgalvao:
> Alternatively, if you think such duplication is unacceptable, I think that only this variable version should be provided.

Let’s do that. Fonts aren’t popular enough to justify duplication. Thank you for your continued effort on this repo.

PR #2096, opened 2020-06-12
switch Powerline fonts to use Git

afeld:

- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.


pelted:
Thanks for this, came to do this when I found this issue was hanging development env deployment. brew/bundle does not handle 404 gracefully on missing repo URLS.

vitorgalvao:
The problem with this approach is that you’re downloading the whole repo for a single font. This doesn’t translate well for humungous repos, we need a better long term solution.

PR #2101, opened 2020-06-14
Provide guidance for OpenType variable fonts

adunning:
When adding the font [Elstob](https://psb1558.github.io/Elstob-font/) (#2078), it emerged that there is no guidance for OpenType variable fonts. These provide more potential functionality, and are fully supported in the latest versions of macOS. A few older applications (notably LaTeX) cannot use them, and need separate files for each weight, but this is being rapidly remedied.

At the moment there are casks providing:

- only variable-width fonts ([Sudo](https://github.com/Homebrew/homebrew-cask-fonts/blob/57fc91532681f3fa7b3a269ede09ac821da8c931/Casks/font-sudo.rb) and [nearly fifty faces from Google Fonts](https://github.com/Homebrew/homebrew-cask-fonts/search?q=wght&unscoped_q=wght));
- only the separate files for different instantiations, even where OpenType variable fonts are available ([Source Code Pro](https://github.com/Homebrew/homebrew-cask-fonts/blob/2ee90f777bff57baac68f725fe4b1008476fc9cb/Casks/font-source-code-pro.rb), [Source Sans Pro](https://github.com/Homebrew/homebrew-cask-fonts/blob/2ee90f777bff57baac68f725fe4b1008476fc9cb/Casks/font-source-sans-pro.rb), [Source Serif Pro](https://github.com/Homebrew/homebrew-cask-fonts/blob/2ee90f777bff57baac68f725fe4b1008476fc9cb/Casks/font-source-serif-pro.rb), [Elstob](https://github.com/Homebrew/homebrew-cask-fonts/blob/f4b55455cdd8d1d2ea9f56edee74aff89c2b6a0d/Casks/font-elstobd.rb)) – corrected in #2077, #2105, #2106;
- both variable fonts and separate files for different weights ([League Spartan](https://github.com/Homebrew/homebrew-cask-fonts/blob/1837305bf78ba5e542595db9c31a382bb44fa394/Casks/font-league-spartan.rb)) – corrected in #2104.

The League Spartan example will result in the system viewing the font as being installed more than once (it will disable one of the copies).

Some developers, however, provide variable font families under a different name. If one installs both the standard OTFs and the variable fonts from the Source Sans Pro distribution, this will provide both 'Source Sans Pro' and 'Source Sans Variable' families, giving both backwards compatibility and access to the full feature set for applications that support it. I would propose that we allow both to be packaged in such circumstances.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,7 +113,7 @@ This constraint may change in the future, when the backend Ruby code becomes mor
 
 ### Multiple Font Formats
 
-If a distribution provides the same font in multiple file formats, for example both OpenType and TrueType, only include one kind. We prefer OTF to TTF, and variable fonts to static files for different instantiations. Where a distribution provides variable and static fonts under different family names, both can be included for backwards compatibility as long as they can be contained within a single cask. For example, [Source Sans](https://github.com/adobe-fonts/source-sans-pro) provides separate 'Source Sans 3' and 'Source Sans 3 VF' families.
+If a distribution provides the same font in multiple file formats, for example both OpenType and TrueType, only include one kind. We prefer OTF to TTF, and variable fonts to static files for different instantiations. Where a distribution provides variable and static fonts under different family names, both can be included for backwards compatibility as long as they can be contained within a single cask. Example: [font-source-serif-pro.rb](https://github.com/Homebrew/homebrew-cask-fonts/blob/ab3e5d7d313b64c0a2c4041196a8eaa8e1539c9c/Casks/font-source-serif-pro.rb#L10#L23).
 
 Note that `font_casker` generates font stanzas for all files, so its output should be edited as needed.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,9 +113,7 @@ This constraint may change in the future, when the backend Ruby code becomes mor
 
 ### Multiple Font Formats
 
-If a distribution provides multiple file formats for the same font, variable fonts are preferred over static files for different instantiations. TTF variable fonts (glyf-based) are more compatible than OTF (CFF2-based). For static instantiations, OTF is preferred to TTF.
-
-Where a distribution provides variable and static fonts under a different family name, both can be included for backwards compatibility. For example, [Source Sans Pro](https://github.com/adobe-fonts/source-sans-pro) provides both 'Source Sans Pro' and 'Source Sans Variable' families.
+If a distribution provides the same font in multiple file formats, for example both OpenType and TrueType, only include one kind. We prefer OTF to TTF, and variable fonts to static files for different instantiations. Where a distribution provides variable and static fonts under different family names, both can be included for backwards compatibility as long as they can be contained within a single formula. For example, [Source Sans](https://github.com/adobe-fonts/source-sans-pro) provides separate 'Source Sans 3' and 'Source Sans 3 VF' families.
 
 Note that `font_casker` generates font stanzas for all files, so its output should be edited as needed.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,9 +113,9 @@ This constraint may change in the future, when the backend Ruby code becomes mor
 
 ### Multiple Font Formats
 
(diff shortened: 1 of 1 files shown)


adunning:
OpenType variable font support was introduced in macOS 10.13. Should the relevant casks be using `depends_on`? (It's not clear to me what happens if one tries to install one of these fonts on an older system.)

khaledhosny:
glyf-based variable fonts (AKA TTF) will show the default instance on older macOS versions (that is a part of the graceful fallback of the design of this format, CFF2-based ones (AKA OTF) will not work at all of not supported, and I think support for them was later than 10.13.

adunning:
Thank you for the explanation! Should TTF versions of OpenType variable fonts be preferred, in that case?

khaledhosny:
I think so, they should work in applications that don’t support font variations (you loss the variations, but at least you will get readable text).

vitorgalvao:
This rule is sensible, and I approve of it and how it’s being handled. @adunning Do as much research as you feel like, revise the rule in this PR, and when you feel confident it’s good ping me back for final tweaks/clarifications and merging.

Three tips:

1. Think *forward*. `10.13` may not seem that long ago, but we’re close to the announcement of `10.16`. Three versions means three years. It’s a long time and most people (and we need to design for “most”) tend to update to the latest version of macOS. Even if that trend was stifled by Catalina, that still leaves a lot of people in the clear (Mojave is `10.14`). Inconveniencing a small number of people now may be better than having to make a rule that will require revisions soon.
2. At the same time, be realistic. If a lot of popular software can’t use variable fonts and won’t be able to for a while, perhaps the technology is indeed too new. That doesn’t seem to be the case, though. If you need some download analytics, I can help with that.
3. Don’t fret too much. Make the best possible decision you can, but it’s not like we’ll hold you to it if it turns out we should have gone another way.

vitorgalvao:
How’s progress here?

vitorgalvao:
Without progress or at least a response to a request for progress report, it makes no sense to keep this open.

adunning:
Might I revive this? My feeling is still that it makes sense to include both variable fonts and the old version where they are installed under different names. LaTeX in particular still shows no sign of getting variable font support, and several other apps specifically look for things under their old names (e.g. BBEdit defaults to Source Code Pro, not Variable).

vitorgalvao:
> My feeling is still that it makes sense to include both variable fonts and the old version where they are installed under different names.

If you mean as different casks, then absolutely not. This repo is already *way* more trouble than it’s worth, and I have no desire to double that effort for no tangible benefit.

> LaTeX in particular still shows no sign of getting variable font support, and several other apps specifically look for things under their old names (e.g. BBEdit defaults to Source Code Pro, not Variable).

I’m seeing an argument to not switch to variable fonts, not an argument to include both.

adunning:
Apart from some holdout programs, variable fonts are far more useful. I am thinking of the case of Source * Pro, which can all be handled within a single cask.

If the repo is more trouble than it's worth, no shame in archiving it, but I've long found it useful!

vitorgalvao:
> I am thinking of the case of Source * Pro, which can all be handled within a single cask.

Are you saying the serif, sans, and code versions could be handled in a single cask? How so?

> If the repo is more trouble than it's worth, no shame in archiving it, but I've long found it useful!

I’m not saying it isn’t useful, I’m saying it’s a bloody mess to keep organised, so we should be as methodical and organised as possible when dealing with it, and automate what we can.

adunning:
No, I am pointing to the distribution of Source Code/Source Serif/Source Sans that puts all the files together in a single download, meaning that one only needs three casks, not six, to have both versions. This does not appear to increase the administrative burden.

vitorgalvao:
> No, I am pointing to the distribution of Source Code/Source Serif/Source Sans that puts all the files together in a single download, meaning that one only needs three casks, not six, to have both versions. This does not appear to increase the administrative burden.

Care to make that change?

adunning:
Source Serif already does this – Source Code done in #2682, Source Sans in #3238.

vitorgalvao:
I’m still not sold that including both is the way to go, but [I did give you leeway on the research](https://github.com/Homebrew/homebrew-cask-fonts/pull/2101#issuecomment-644231850), so I’ll trust and merge this for now as the accepted rule.

Thank you.

vitorgalvao (approved):

PR #2098, opened 2020-06-12
add/update all Google Fonts, take II

afeld:
Follow-up to https://github.com/Homebrew/homebrew-cask-fonts/pull/2095, with the requested changes. Squashed most of the commits for readability. Since the diff is so huge, here is the link to [the script](https://github.com/afeld/homebrew-cask-fonts/blob/google-fonts-B/developer/bin/import_google_fonts) and [the template](https://github.com/afeld/homebrew-cask-fonts/blob/google-fonts-B/developer/cask.rb.j2) by themselves.

This pull request adds 548 Google font casks, deletes five missing ones, and fixes dozens of others.

---

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` reports no offenses.
- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.
- [x] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

Additionally, **if adding a new cask**:

- [x] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.
- [x] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=Issues).
- [x] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).

PR #2108, opened 2020-06-14
move Google Fonts cask template into script

afeld:
As requested in https://github.com/Homebrew/homebrew-cask-fonts/pull/2098#issuecomment-643755236.


vitorgalvao:
Thank you!

PR #2111, opened 2020-06-14
Lora: Use upstream repository

adunning:

- [X] `brew cask audit --download {{cask_file}}` is error-free.
- [X] `brew cask style --fix {{cask_file}}` reports no offenses.
- [X] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.
- [X] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

### DIFF
--- a/Casks/font-lora.rb
+++ b/Casks/font-lora.rb
@@ -1,16 +1,12 @@
 cask 'font-lora' do
-  version :latest
-  sha256 :no_check
+  version '2.210'
+  sha256 'c872190b80eedbc72bf0a84b56e877cdf5eefae4377574df023b1e8b787ab65a'
 
-  # github.com/google/fonts/ was verified as official when first introduced to the cask
-  url 'https://github.com/google/fonts/trunk/ofl/lora',
-      using:      :svn,
-      trust_cert: true
+  url "https://github.com/cyrealtype/Lora-Cyrillic/releases/download/v#{version}/Lora-Variable.zip"
+  appcast 'https://github.com/cyrealtype/Lora-Cyrillic/releases.atom'
   name 'Lora'
-  homepage 'https://fonts.google.com/specimen/Lora'
+  homepage 'https://github.com/cyrealtype/Lora-Cyrillic'
 
-  depends_on macos: '>= :sierra'
-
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but we just got [a script to keep all Google Fonts updated](https://github.com/Homebrew/homebrew-cask-fonts/pull/2098), so for the moment it’s a better idea to keep those when they exist than to fragment to different repos.

adunning:
`CONTRIBUTING.md` encourages users to switch to upstream repositories where available – does that no longer apply? If we're using unversioned sources instead, how would Homebrew know to update an installed font?

vitorgalvao:
> `CONTRIBUTING.md` encourages users to switch to upstream repositories where available – does that no longer apply?

It’s been a long time since we’ve made that rule. In general I think it’s a good one, but with the addition of the script I also think we should change it.

In truth, this repo requires more maintenance than it’s worth. Most fonts have too few users, but unlike other repos, removing them for low usage seems less warranted. That leads to broken font casks over time, which despite being broken for months or years are only noticed occasionally by users.

Having an automated way to ensure those casks are always working is a huge boon to maintainability. I don’t really like that we’re relying on Google for it ([I wonder when they’ll get tired of it](https://killedbygoogle.com/)) but it’s not like we have a big choice—most of those can only be retrieved from Google Fonts. Whatever we do, we’d have to do a lot of reworking on this repo if Google Fonts went away. I’d see that as a boon, even, clean up everything and readd only what people still want.

> If we're using unversioned sources instead, how would Homebrew know to update an installed font?

It wouldn’t, that’s true. But it’s not like most fonts are updated often (or at all) or are problematic if on an older version, or even have a version.

In general, relying more on Google Fonts seems to be the choice that keeps this tap healthier, and it needs that help.

vitorgalvao:
https://github.com/Homebrew/homebrew-cask-fonts/pull/2113

PR #2182, opened 2020-07-01
Update font-cascadia from 2005.15 to 2007.01

peterlewis:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

After making all changes to a cask, verify:

- [x] `brew cask audit --download {{cask_file}}` is error-free.
- [x] `brew cask style --fix {{cask_file}}` left no offenses.
- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask/pulls) for the same update.
- [ ] The submission is for [a stable version](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#stable-versions) or [documented exception](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#but-there-is-no-stable-version).

PR #2183, opened 2020-07-01
Update font-cascadia variants from 2005.15 to 2007.01

sal0max:

- [ ] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.

There are open pull requests:
- #2181 which fails
- #2182 which only updates `font-cascadia`, but lacks the derivates `font-cascadia-mono`, `font-cascadia-mono-pl` and `font-cascadia-pl`

### DIFF
--- a/Casks/font-cascadia-mono-pl.rb
+++ b/Casks/font-cascadia-mono-pl.rb
@@ -2,7 +2,7 @@ cask 'font-cascadia-mono-pl' do
   version '2007.01'
   sha256 '9f066d0d9cb2669bea2e130d7add43d496bf24ef995f42dc603fc2014574a3a4'
 
-  url "https://github.com/microsoft/cascadia-code/releases/download/v#{version}/CascadiaCode_#{version}.zip"
+  url "https://github.com/microsoft/cascadia-code/releases/download/v#{version}/CascadiaCode-#{version}.zip"
   appcast 'https://github.com/microsoft/cascadia-code/releases.atom'
   name 'Cascadia Mono PL'
   homepage 'https://github.com/microsoft/cascadia-code'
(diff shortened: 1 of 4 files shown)


peterlewis:
Ah, I'm really sorry @sal0max. I thought "ooh, I'll do the variants too" in the PR, but I think that in the meantime you kindly did this!

So, if you're happy to add the standard to your PR I'll close mine? Else if you'd prefer then I believe mine is all hunky dory (albeit now a poorly-named PR).

peterlewis:
Ah interesting. I've just noticed that rather than including the different weights (and sticking with OTF), you've gone for the TTF (which includes all weights in a single file).

I'm no expert in fonts, so don't know the ins-and-outs. Is there a reason for that?

sal0max:
Switched to ttf because technically, there should be no difference in usage. And by using just one ttf instead of several, I thought it would be more fail-safe in the future. But tbh I don't really have a preference.

Concerning which pull request to merge, maybe let the maintainers decide. I thought you missed the derivates, so I made this. I'm fine with either one.

peterlewis:
> Switched to ttf because technically, there should be no difference in usage. And by using just one ttf instead of several, I thought it would be more fail-safe in the future. But tbh I don't really have a preference.
> 
> Concerning which pull request to merge, maybe let the maintainers decide. I thought you missed the derivates, so I made this. I'm fine with either one.

Hah, indeed I did miss the variants, which we then likely spotted at the same time! :) Cool beans, happy to let the maintainers decide which is more suitable.

vitorgalvao:
Variable fonts seem to be relatively well supported these days, so let’s go with that.

PR #2249, opened 2020-08-05
automatically import Google Fonts

afeld:
Rather than have to run the Google Font updates manually, this pull request adds a GitHub Actions workflow to do it automatically. You can see the example runs [here](https://github.com/afeld/homebrew-cask-fonts/actions?query=branch%3Aauto-import-google+workflow%3A%22Update+Google+Fonts%22).

### DIFF
--- a/.github/workflows/google_fonts.yml
+++ b/.github/workflows/google_fonts.yml
@@ -3,7 +3,7 @@ name: Update Google Fonts
 on:
   push: # TODO remove
   schedule:
-    - cron: "0 0 * * *" # daily
+    - cron: "0 0 * * 0" # weekly
 
 jobs:
   update:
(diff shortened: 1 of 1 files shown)


.github/workflows/google_fonts.yml
```diff
@@ -0,0 +1,34 @@
+name: Update Google Fonts
+
+on:
+  push: # TODO remove
```
afeld, line 4: Just put this in there for demonstration purposes in this pull request; will remove before (or right after) merge.

.github/workflows/google_fonts.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: pip3 install gftools jinja2 protobuf
+
+      - name: Import Google Fonts
+        run: ./developer/bin/import_google_fonts vendor/google-fonts
+
+      - name: Commit updated files
+        uses: EndBug/add-and-commit@v4
+        with:
+          add: Casks
+          message: automatic import of Google Fonts from GitHub Actions
```
afeld, line 34: This will commit the change on the primary branch (`master`). Can change it to submit a pull request if that's preferable, but I figured it's not worth the noise.

afeld: On the other hand, might be nice to ensure that the various CI checks are run to make sure the updates conform. Thoughts?

vitorgalvao: > Can change it to submit a pull request if that's preferable

It is. As you mentioned, it’ll ensure CI passes and will let us catch something that may be going wrong. Merging a good PR is not much trouble.

vitorgalvao:
Pinging @reitermarkus in case you’re interested in reviewing the GitHub actions part.

.github/workflows/google_fonts.yml
afeld: Done.

afeld:
Ok, updated to send a pull request rather than committing directly to `master`. Example: https://github.com/afeld/homebrew-cask-fonts/pull/2

.github/workflows/google_fonts.yml
afeld:
```suggestion
```

vitorgalvao:
Thank you! This will greatly ease the burden of this repo.

Issue #2255, opened 2020-08-13
duplicate casks for ubuntu font family


zhangyoufu:
#### Description of issue

`font-ubuntu` contains `Ubuntu Mono` and `Ubuntu Condensed`. `font-ubuntu-mono` & `font-ubuntu-condensed` are duplicated. (introduced in #2098)

vitorgalvao:
https://github.com/Homebrew/homebrew-cask-fonts/pull/2257

PR #2257, opened 2020-08-14
Delete font-ubuntu.rb

vitorgalvao:
Refs https://github.com/Homebrew/homebrew-cask-fonts/issues/2255.

This is already covered by the fonts added automatically by the Google Fonts repos script.


zhangyoufu:
Google font should not be considered as the official upstream of Ubuntu font family. Google’s repo does not have latest copy of Ubuntu font family.

vitorgalvao:
> Google font should not be considered as the official upstream of Ubuntu font family. Google’s repo does not have latest copy of Ubuntu font family.

Unfortunately, this repo is way more trouble than it’s worth, so we’re trying to cut work to a minimum. everything we can automate, we do, and the Google Fons repo is automatable.

If that means a few fonts have to be a little behind, so be it. If you want to help, the correct solution is to open an issue or make a PR on the Google Fonts repo to fix what’s missing. That way it’s fixed for far more people.

zhangyoufu:
I agree that things should be automated. But I don’t think everything relying on Google’s font repo is the correct way to go. I would stop using cask-fonts if this happens.

vitorgalvao:
> I would stop using cask-fonts if this happens.

That’s your choice. It has no effect on this decision.

PR #2263, opened 2020-08-15
Improve automatic import of Google Fonts.

reitermarkus:
Automatically guess descriptions.


reitermarkus:
cc @afeld

developer/bin/import_google_fonts
```diff
@@ -55,23 +58,31 @@ end
...
         self.early_access = early_access
 
     def font_name(self):
         return self.meta.name
 
+    def description(self):
+        if not self.desc:
+          return None
+
+        if len(self.desc) == 0:
```
afeld, line 74: Minor: This would be caught by the check above, as empty strings in Python are falsy.

PR #2394, opened 2020-09-14
font-437-plus: New Cask - Ultimate Oldschool PC Font Pack v2.0

Kreeblah:
Note: This is based on an existing cask named font-px437-pxplus.  I created a new cask instead of updating the old one, as the updates to the fonts are noted as incompatible with the original version per the note at https://int10h.org/oldschool-pc-fonts/readme/#history  I also added a conflicts_with statement referencing the original cask.

- [x] There are no [open pull requests](https://github.com/Homebrew/homebrew-cask-fonts/pulls) for the same update.

- [x] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] `brew cask audit --new-cask {{cask_file}}` worked successfully.
- [x] `brew cask install {{cask_file}}` worked successfully.
- [x] `brew cask uninstall {{cask_file}}` worked successfully.
- [x] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).


reitermarkus:
> fonts are noted as incompatible with the original version

I guess that only is an issue if you install both versions side-by-side. Not sure if we should keep the old version at all then.

Casks/font-437-plus.rb
```diff
@@ -0,0 +1,743 @@
+cask "font-437-plus" do
+  version "2.0"
+  sha256 "e2130bc9de9d2362275e4cb9cbd927bc0cc3dd10c14a2c5a661ae32cd5d585aa"
+
+  url "https://int10h.org/oldschool-pc-fonts/download/oldschool_pc_font_pack_v#{version}_FULL.zip"
```
reitermarkus, line 5: Let's use the `https://int10h.org/oldschool-pc-fonts/download/oldschool_pc_font_pack_v2.0_ttf.zip` URL since we only install the TTF files anyways.

Kreeblah: Sure.  I just switched over to that and pushed another update.

Kreeblah:
> > fonts are noted as incompatible with the original version
> 
> I guess that only is an issue if you install both versions side-by-side. Not sure if we should keep the old version at all then.

The only reason I can think of to keep the old one around would be in case people have configured something to depend on one of the fonts from it that isn't in the new version.  If that's not a concern, though, it doesn't seem like it'd be too worthwhile to keep the old one.

vitorgalvao:
> Not sure if we should keep the old version at all then.

[I don’t think we should](https://github.com/Homebrew/homebrew-cask-fonts/pull/2396).

Regarding the naming of this cask, we could go one of three ways:

1. `font-437-plus` (current).
2. `font-oldschool-pc` (more accurate reflection of the pack).
3. Separate each font into their own cask.

I don’t feel strongly about it but think we should go with the **2** or **3**.

reitermarkus:
@vitorgalvao, I think `font-437-plus` is fine since all font names include “437“ or “plus”.

vitorgalvao (approved):

Issue #3867, opened 2021-04-10
Follow the naming policy of the upstream project


heavywatal:
Like it or not, ["Source Sans Pro" has been renamed to "Source Sans 3"](https://github.com/adobe-fonts/source-sans/issues/192). According to this decision, `font-source-sans-pro` should stick to "Source Sans Pro v2.xxx", and a new formula `font-source-sans-3` (or `font-source-sans@3`?) should be added to track "Source Sans 3".  [Source Serif 4](https://github.com/adobe-fonts/source-serif/issues/77) likewise.

vitorgalvao:
Please submit a PR so we can properly discuss it.

PR #3898, opened 2021-04-16
Source fonts naming

heavywatal:
This PR addresses the issue #3867.
The following changes have been made.

- Source Sans 3: new cask
- Source Sans Pro: downgrade to version 2
- Source Serif 4: new cask
- Source Serif Pro: update file source and point to version 3

- [x] `brew audit --cask {{cask_file}}` is error-free.
- [x] `brew style --fix {{cask_file}}` reports no offenses.

- [x] Named the cask according to the [token reference](https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/token_reference.md).
- [x] Checked the cask is submitted to [the correct repo](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask).
- [x] `brew audit --new-cask {{cask_file}}` worked successfully.
- [x] `brew install --cask {{cask_file}}` worked successfully.
- [x] `brew uninstall --cask {{cask_file}}` worked successfully.


Loyalsoldier:
I think it's better to do like this:

Source Sans: new cask
Source Sans Pro: downgrade to version 2
Source Serif
Source Serif Pro: update file source and point to version 3

And use `caveats` to warn users about the newer release of casks like https://github.com/Homebrew/homebrew-cask-fonts/pull/3900

And deprecate `Source Sans Pro` and `Source Serif Pro` in a year or so.

vitorgalvao:
What’s the point in keeping the old versions? They’re even failing `livecheck`, which should indicate the new ones are the continuation.

Casks/font-source-sans-3.rb
```diff
@@ -0,0 +1,30 @@
+cask "font-source-sans-3" do
```
vitorgalvao, line 1: Versions shouldn’t be part of the name.

Casks/font-source-sans-3.rb
```diff
@@ -0,0 +1,30 @@
...
+  version "3.028"
+  sha256 "1186d594ecb1336679a1d102bb6ddbe6041f0fd029482e53f02e46aa3ab011cf"
+
+  url "https://github.com/adobe-fonts/source-sans/releases/download/#{version}R/source-sans-#{version.sub(/\./, "v")}R.zip"
```
vitorgalvao, line 5:
```suggestion
  url "https://github.com/adobe-fonts/source-sans/releases/download/#{version}R/source-sans-#{version.major}v#{version.minor}R.zip"
```

Casks/font-source-sans-3.rb
```diff
@@ -0,0 +1,30 @@
...
+  name "Source Sans 3"
```
vitorgalvao, line 6: No version in name.

Casks/font-source-sans-3.rb
```diff
@@ -0,0 +1,30 @@
...
+  desc "Fonts designed for user interfaces"
+  homepage "https://github.com/adobe-fonts/source-sans"
+
+  livecheck do
+    url "https://github.com/adobe-fonts/source-sans/releases/latest"
+    strategy :page_match
+    regex(%r{tag/(\d+(?:\.\d+)*)}i)
+  end
+
+  font "OTF/SourceSans3-Black.otf"
```
vitorgalvao, line 16:
```suggestion
  font "OTF/SourceSans#{version.major}-Black.otf"
```

In all of them.

vitorgalvao (changes requested):

heavywatal:
Thank you for your feedback. I will consider `caveats` and the other requested changes after finishing this discussion.

But first, please read the upstream issues linked in #3867 again. "Source Sans 3" is not just the version 3 of "Source Sans", but it is the new family name. There is no font family named "Source Sans". Besides, the new one seems not geometrically upward-compatible with old "Source Sans Pro". In other words, silent upgrading from "Source Sans Pro" to "Source Sans 3" (and removing old "Source Sans Pro") will break existing documents. And it is expected to happen again when a new font family like "Source Sans 4" is released.

(I personally don't like their decision, but) keeping old fonts will be necessary to protect users' environments and documents.  Now I have two questions:

1. Based on the fact that "Source Sans 3" is the family name, should the cask be named `source-sans-3` or `source-sans`( and `source-sans@3` in the future)?
2. Is it possible to bypass `livecheck` for casks without `livecheck` entry?

Loyalsoldier:
Is there any alias in this repo as the ones in `Homebrew/homebrew-core`?

Is so, the old `source sans pro` can then be linked to `source sans`, and `source serif pro` be linked to `source serif`.

vitorgalvao:
> "Source Sans 3" is not just the version 3 of "Source Sans", but it is the new family name.

I mean, kind of. The repo doesn’t have the number in the name, nor does the first sentence in the README.

image

> (I personally don't like their decision, but)

Neither do I, but coming from Adobe I’m not surprised. We’ve removed plenty of their software from Homebrew Cask for user-hostility, and I’m inclined to do the same for these fonts.

[Our current preferred way to add fonts to this repo is via Google Fonts](https://github.com/Homebrew/homebrew-cask-fonts/blob/master/CONTRIBUTING.md#when-to-contribute), and I see they do have the Pro version. So it’s probably best to get rid of these and let Google Fonts deal with the naming however deem best; we’ll follow it automatically.

> 1. Based on the fact that "Source Sans 3" is the family name, should the cask be named `source-sans-3` or `source-sans`( and `source-sans@3` in the future)?

We don’t do the `@version` for casks and definitely don’t keep different versions of fonts. They already are more trouble than they’re worth without that.

> 2\. Is it possible to bypass `livecheck` for casks without `livecheck` entry?

CI always runs the check.

Sorry (and thank you) for the spent effort for a refusal, but it was necessary to see this in context to understand the situation. You can still use the fonts in your personal tap, though.

Will leave it open for a bit in case there are other ideas. Or perhaps they’ll come to their senses (though I doubt it).

heavywatal:
OK, agree that they should be automatically managed via Google Fonts. And now I understand that each cask name should correspond to its distribution, not fonts in it. Thank you for reviewing the issue and PR.

FYI, by "family name" I meant the attribute of the font files (how it is referenced in an OS):
Screen Shot 2021-04-18 at 0 44 17

PR #3903, opened 2021-04-17
Remove fonts from abobe-fonts repos

vitorgalvao:
Tired of Adobe to no end; they can’t even be normal with fonts. See https://github.com/Homebrew/homebrew-cask-fonts/pull/3898#issuecomment-821838947.

This PR removes all casks from `github.com/adobe-fonts`. **That doesn’t mean the fonts will become unavailable**—because they are part of the Google Fonts repo, they will be added again via our automated tooling. This simply means less effort for us (and our users) that we won’t have to deal with the (seemingly universally disliked) kerfuffle of deciding what to do with fonts with versions in the name.

Closes https://github.com/Homebrew/homebrew-cask-fonts/pull/3898. Let’s merge this after May 10, to give time for other ideas.


vitorgalvao:
Keeps getting conflicts. I’ll do one more and merge right away.

allenchen10:
I don't think Source Han fonts are part of the Google Fonts repo. They are similar to Noto CJK fonts but vary in weights. Maybe we can keep them for now?

vitorgalvao:
> I don't think Source Han fonts are part of the Google Fonts repo.

Then you should [submit them](https://github.com/Homebrew/homebrew-cask-fonts/blob/master/CONTRIBUTING.md#when-to-contribute). That’s the only way they’re coming back to this repo right now.

iandol:
@victorgalvao -- this is just to comment: the versions from Google are ancient. I just had to redo all my dependencies changing the names when Adobe changed the name, annoying but do-able. But now suddenly the names have all gone back AND I have old versions of fonts with less glyphs (Google has the V2.02 version from 2015, there have been LOTS of changes to these fonts since then! -- https://github.com/adobe-fonts/source-sans/releases -- for example greek italics available since V3.06) and this regression in rendering and glyph coverage has a much bigger negative impact than the name change.

Google do not seem to update their fonts reliably: https://github.com/google/fonts/issues/1744

Thank you as always for the work you do in maintaining brew, I do appreciate that you must factor in maintenance and support burdens...

vitorgalvao:
@iandol Thank you for the kind words and understanding.

Unfortunately, this repo is way more trouble than it’s worth when it requires manual changes, and Adobe in particular has burned us way too often. In the grand scheme of their user hostile history, this was by far the tamest (of the ones affecting HBC). But it also shows there’s no area of software where they won’t be problematic.

We’ll continue to support the remaining Adobe software in casks for as long as the current solutions work, but I’ll have no reservations about removing them on the next imbecilic change they make. I’d rather not support a company which doesn’t respect users than have maintainers burnout.

Issue #2239, opened 2020-07-31
Any reason to keep on subversion?


gcbw:
Before i send in PRs, is there any reason to keep most fonts on subversion instead of git directly?

e.g. https://github.com/Homebrew/homebrew-cask-fonts/blob/master/Casks/font-ubuntu-mono.rb

vitorgalvao:
Subversion allows us to download only what we need. Without it, every time a user wanted a Google Fonts font, they’d have to download the whole repo.

gcbw:
```

$ git clone --depth 1 --no-checkout --filter=blob:none https://github.com/google/fonts
Cloning into 'fonts'...
remote: Enumerating objects: 1367, done.
remote: Counting objects: 100% (1367/1367), done.
remote: Compressing objects: 100% (1338/1338), done.
remote: Total 1367 (delta 0), reused 611 (delta 0), pack-reused 0
Receiving objects: 100% (1367/1367), 332.11 KiB | 1.42 MiB/s, done.

$ cd fonts

$ git checkout master -- ufl/ubuntumono
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 510 bytes | 510.00 KiB/s, done.
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 1.98 KiB | 1.98 MiB/s, done.
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 696 bytes | 696.00 KiB/s, done.
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 4.27 KiB | 4.27 MiB/s, done.
remote: Enumerating objects: 1, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 1 (delta 0), reused 1 (delta 0), pack-reused 0
Receiving objects: 100% (1/1), 3.69 KiB | 3.69 MiB/s, done.
remote: Enumerating objects: 1, done.
[...]

PR #4055, opened 2021-05-17
Don't use SVN to clone a GitHub repo

verhovsky:
Currently brew fails to update the font and says "You must: brew install svn". The pull request that made this change added this to other formulas, so it seems like the bot needs to be fixed.

### DIFF
--- a/Casks/font-source-code-pro.rb
+++ b/Casks/font-source-code-pro.rb
@@ -3,8 +3,7 @@ cask "font-source-code-pro" do
   sha256 :no_check
 
   url "https://github.com/google/fonts/trunk/ofl/sourcecodepro",
-      verified: "github.com/google/fonts/",
-      using:    :svn
+      verified: "github.com/google/fonts/"
   name "Source Code Pro"
   homepage "https://fonts.google.com/specimen/Source+Code+Pro"
 


vitorgalvao:
This is necessary because git can’t just download a directory from a repo, but svn can. Without this, you’d need to clone the entire Google Fonts repo to install a single font.

verhovsky:
@vitorgalvao thanks for explaining it to me. Just so you know, you can download files from GitHub directly, for example this is how I install Roboto Mono on Ubuntu:

```sh
sudo wget --content-disposition -P /usr/share/fonts/truetype/robotomono https://github.com/google/fonts/blob/master/apache/robotomono/static/RobotoMono-{Bold,BoldItalic,Italic,Light,LightItalic,Medium,MediumItalic,Regular,Thin,ThinItalic}.ttf?raw=true
```

but I guess it's probably easier with svn.

duraki:
I agree w/ @verhovsky, indeed it might be easier to use `svn` pkg, but lets be honest; installing an `svn` just so I could support few other fonts seems quite tedious. Maybe a "prettier" UX is expected in my opinion to contribute to this, a la:

```sh
This cask requires 'svn' for downloading a part of Git directory. Do you want to install it automatically [Y/n]: y
# => brew install svn
```

Adding a pop-up or default* `svn` uninstall add the end of the installation would be cool (can't assure it's possible w/ brew natively?):

```sh
Do you want to automatically uninstall 'svn' as a part of dependecies [Y/n]: 
# => brew uninstall svn
```

I'm saying it loudly because [this is](https://github.com/Homebrew/homebrew-cask-fonts/issues?q=svn) probably some kind of metric, no?

---

@verhovsky: `wget` is awesome no doubt, but try playing a bit with your term/teach environment:

```sh
$ echo "alias wget=\"echo 'No, dude: curl [http://example.com] --output my.file'\"" >> ~/.zshrc && source ~/.zshrc
$ wget http://example.com
# => No, dude: curl [http://example.com] --output my.file
...

# => after a few days revisit your bashrc for regular cleaning
```

Btw. Another [task to ref](https://github.com/Homebrew/homebrew-cask-fonts/issues/2239) for Repository Owner and Authors' opinion, which realistically makes sense.

vitorgalvao:
> indeed it might be easier to use `svn` pkg

You’re not paying attention. Read what I wrote:

> It’s not that it’s easier with SVN, it’s that it’s the only reasonable way given the constraints we have

Frankly, this repo isn’t worth the maintainer time to do something different. There are plenty of urgent things sorely lacking from HBC.

> but lets be honest; installing an svn just so I could support few other fonts seems quite tedious.

As arguments for not requiring `svn` go, that’s the least compelling I remember reading. Having to run `brew install svn` *once* would have been less tedious and faster than what it took you to write that sentence, or I to read it.

> Adding a pop-up or default* svn uninstall add the end of the installation would be cool (can't assure it's possible w/ brew natively?)

You’re thinking of what you want, not what’s best for most users. Installing and removing a package all the time does not make sense. It wastes time, bandwidth, electricity, disk wear…

> I'm saying it loudly because [this is](https://github.com/Homebrew/homebrew-cask-fonts/issues?q=svn) probably some kind of metric, no?

No. What’s the shock that issues are referencing a dependency?

If anything, it might be a metric of people not searching before posting.

> Btw. Another [task to ref](https://github.com/Homebrew/homebrew-cask-fonts/issues/2239) for Repository Owner and Authors' opinion, which realistically makes sense.

Please stop posting closed issues without context. **Again, read what’s been posted**. What’s your point? That issue has been has been addressed as not feasible right now.

Either offer a concrete solution, preferably in the form of a PR, or asking us to reread old closed issues is a waste of time.

vitorgalvao:
> indeed it might be easier to use `svn` pkg

I’ve addressed that in my previous comment:

> It’s not that it’s easier with SVN, it’s that it’s the only reasonable way given the constraints we have

“Easier” isn’t the reason we do it this way.

> installing an svn just so I could support few other fonts seems quite tedious.

Again, *we don’t have a better way*. Having to run `brew install svn` *once* isn’t a large enough problem, nor would I classify it as “tedious”. But if you think it is and come up with a concrete solution, we will consider a PR.

> Adding a pop-up or default* svn uninstall add the end of the installation would be cool (can't assure it's possible w/ brew natively?)

You’re thinking of what you want, not what’s best for most users. Installing and removing a package all the time would waste time, bandwidth, electricity, disk wear…

And who would want a pop-up? Interrupting a task for such a trivial reason would go against HB’s philosophy.

> I'm saying it loudly because [this is](https://github.com/Homebrew/homebrew-cask-fonts/issues?q=svn) probably some kind of metric, no?

If anything, it might be a metric of people not searching before posting.

Otherwise, what’s the shock that issues are referencing a dependency?

PR #4760, opened 2021-08-01
Add version constraints

sxyazi:

- [ ] `brew audit --cask <cask>` is error-free.

### DIFF
--- a/Casks/font-source-code-pro.rb
+++ b/Casks/font-source-code-pro.rb
@@ -1,5 +1,5 @@
 cask "font-source-code-pro" do
-  version :latest
+  version "2.030"
   sha256 :no_check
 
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Thank you, but casks form Google Fonts are handled by a script.

sxyazi:
@vitorgalvao I found an issue where the font always gets updated when I run `brew upgrade --greedy`, but the others don't, so I suspect it has something to do with the version.

bevanjkay:
> @vitorgalvao I found an issue where the font always gets updated when I run `brew upgrade --greedy`, but the others don't, so I suspect it has something to do with the version.

This is the expected behaviour. 
Any cask with `version :latest` will be upgraded when you pass the `--greedy` flag. 

If the casks are versioned they are only upgraded when a newer version is available.

PR #5758, opened 2022-05-04
Use git instead of svn for IBM Plex Mono

hmarr:

---

Hello!

This is a draft PR to demonstrate how we could use the proposed support for git partial clones with sparse checkouts as an alternative to subversion, as [suggested](https://github.com/Homebrew/brew/pull/13232#issuecomment-1115868261) by @MikeMcQuaid.

Using this branch with the master branch of homebrew (which includes https://github.com/Homebrew/brew/pull/14035) shows that we retain the speed benefits of not fetching the full fonts repo:

```
~ $ time brew install --cask font-ibm-plex-mono
==> Cloning https://github.com/google/fonts.git
Cloning into '/Users/hmarr/Library/Caches/Homebrew/Cask/font-ibm-plex-mono--git-sparse'...
==> Checking out branch main
Already on 'main'
Your branch is up to date with 'origin/main'.
Warning: No checksum defined for cask 'font-ibm-plex-mono', skipping verification.
==> Installing Cask font-ibm-plex-mono
==> Moving Font 'IBMPlexMono-ThinItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-ThinItalic
==> Moving Font 'IBMPlexMono-BoldItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-BoldItalic
==> Moving Font 'IBMPlexMono-ExtraLight.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-ExtraLight
==> Moving Font 'IBMPlexMono-ExtraLightItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Extr
==> Moving Font 'IBMPlexMono-Italic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Italic.ttf'
==> Moving Font 'IBMPlexMono-Light.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Light.ttf'
==> Moving Font 'IBMPlexMono-LightItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-LightItal
==> Moving Font 'IBMPlexMono-Medium.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Medium.ttf'
==> Moving Font 'IBMPlexMono-MediumItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-MediumIt
==> Moving Font 'IBMPlexMono-Regular.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Regular.ttf'
==> Moving Font 'IBMPlexMono-SemiBold.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-SemiBold.ttf
==> Moving Font 'IBMPlexMono-SemiBoldItalic.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-SemiBo
==> Moving Font 'IBMPlexMono-Thin.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Thin.ttf'
==> Moving Font 'IBMPlexMono-Bold.ttf' to '/Users/hmarr/Library/Fonts/IBMPlexMono-Bold.ttf'
🍺  font-ibm-plex-mono was successfully installed!
brew install --cask font-ibm-plex-mono  5.31s user 2.14s system 79% cpu 9.369 total
```

### DIFF
--- a/Casks/font-ibm-plex-mono.rb
+++ b/Casks/font-ibm-plex-mono.rb
@@ -2,11 +2,10 @@ cask "font-ibm-plex-mono" do
   version :latest
   sha256 :no_check
 
-  url "https://github.com/google/fonts",
+  url "https://github.com/google/fonts.git",
       verified:   "github.com/google/fonts",
       branch:     "main",
-      only_paths: ["ofl/ibmplexmono"],
-      using:      :git
+      only_paths: ["ofl/ibmplexmono"]
   name "IBM Plex Mono"
   homepage "https://fonts.google.com/specimen/IBM+Plex+Mono"
 
--- a/Casks/font-ibm-plex-mono.rb
+++ b/Casks/font-ibm-plex-mono.rb
@@ -3,24 +3,24 @@ cask "font-ibm-plex-mono" do
   sha256 :no_check
 
   url "https://github.com/google/fonts.git",
-      verified:   "github.com/google/fonts",
-      branch:     "main",
-      only_paths: ["ofl/ibmplexmono"]
+      verified:  "github.com/google/fonts",
+      branch:    "main",
+      only_path: "ofl/ibmplexmono"
   name "IBM Plex Mono"
   homepage "https://fonts.google.com/specimen/IBM+Plex+Mono"
 
-  font "ofl/ibmplexmono/IBMPlexMono-Bold.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-BoldItalic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-ExtraLight.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-ExtraLightItalic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-Italic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-Light.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-LightItalic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-Medium.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-MediumItalic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-Regular.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-SemiBold.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-SemiBoldItalic.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-Thin.ttf"
-  font "ofl/ibmplexmono/IBMPlexMono-ThinItalic.ttf"
+  font "IBMPlexMono-Bold.ttf"
+  font "IBMPlexMono-BoldItalic.ttf"
+  font "IBMPlexMono-ExtraLight.ttf"
+  font "IBMPlexMono-ExtraLightItalic.ttf"
+  font "IBMPlexMono-Italic.ttf"
+  font "IBMPlexMono-Light.ttf"
+  font "IBMPlexMono-LightItalic.ttf"
+  font "IBMPlexMono-Medium.ttf"
+  font "IBMPlexMono-MediumItalic.ttf"
+  font "IBMPlexMono-Regular.ttf"
+  font "IBMPlexMono-SemiBold.ttf"
+  font "IBMPlexMono-SemiBoldItalic.ttf"
+  font "IBMPlexMono-Thin.ttf"
+  font "IBMPlexMono-ThinItalic.ttf"
 end
--- a/Casks/font-ibm-plex-mono.rb
+++ b/Casks/font-ibm-plex-mono.rb
@@ -7,6 +7,7 @@ cask "font-ibm-plex-mono" do
       branch:    "main",
       only_path: "ofl/ibmplexmono"
   name "IBM Plex Mono"
+  desc "Corporate typeface for IBM"
   homepage "https://fonts.google.com/specimen/IBM+Plex+Mono"
 
   font "IBMPlexMono-Bold.ttf"
--- a/Casks/font-ibm-plex-mono.rb
+++ b/Casks/font-ibm-plex-mono.rb
@@ -2,24 +2,26 @@ cask "font-ibm-plex-mono" do
   version :latest
   sha256 :no_check
 
-  url "https://github.com/google/fonts/trunk/ofl/ibmplexmono",
-      verified: "github.com/google/fonts/",
-      using:    :svn
+  url "https://github.com/google/fonts",
+      verified:   "github.com/google/fonts",
+      branch:     "main",
+      only_paths: ["ofl/ibmplexmono"],
+      using:      :git
   name "IBM Plex Mono"
   homepage "https://fonts.google.com/specimen/IBM+Plex+Mono"
 
-  font "IBMPlexMono-Bold.ttf"
-  font "IBMPlexMono-BoldItalic.ttf"
-  font "IBMPlexMono-ExtraLight.ttf"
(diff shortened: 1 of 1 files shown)


Casks/font-ibm-plex-mono.rb
```diff
@@ -2,24 +2,26 @@
...
   sha256 :no_check
 
-  url "https://github.com/google/fonts/trunk/ofl/ibmplexmono",
-      verified: "github.com/google/fonts/",
-      using:    :svn
+  url "https://github.com/google/fonts",
+      verified:   "github.com/google/fonts",
+      branch:     "main",
+      only_paths: ["ofl/ibmplexmono"],
+      using:      :git
```
MikeMcQuaid, line 9:
```suggestion
      only_paths: ["ofl/ibmplexmono"]
```

This might even work (or omitting just one of them). No idea but might be worth playing with.

MikeMcQuaid (dismissed):

Casks/font-ibm-plex-mono.rb
hmarr: Ah, good point! I saw that the download strategy detector was choosing the `GitHubGitDownloadStrategy`, but I hadn't registered that it's a subclass of `GitDownloadStrategy`. I confirmed that removing `using: :git` and instead appending `.git` to the URL does work 👍 

I tried removing `branch`, but it seems like we assume `master` is the default branch:

```
==> Cloning https://github.com/google/fonts.git
Cloning into '/Users/hmarr/Library/Caches/Homebrew/Cask/font-ibm-plex-mono--git'...
fatal: Remote branch master not found in upstream origin
Error: Download failed on Cask 'font-ibm-plex-mono' with message: Failure while executing; `git clone --branch master --no-checkout --filter=blob:none -c advice.detachedHead=false https://github.com/google/fonts.git /Users/hmarr/Library/Caches/Homebrew/Cask/font-ibm-plex-mono--git` exited with 128. Here's the output:
Cloning into '/Users/hmarr/Library/Caches/Homebrew/Cask/font-ibm-plex-mono--git'...
fatal: Remote branch master not found in upstream origin
```

Casks/font-ibm-plex-mono.rb
```diff
@@ -2,24 +2,25 @@
   version :latest
   sha256 :no_check
 
-  url "https://github.com/google/fonts/trunk/ofl/ibmplexmono",
-      verified: "github.com/google/fonts/",
-      using:    :svn
+  url "https://github.com/google/fonts.git",
+      verified:   "github.com/google/fonts",
+      branch:     "main",
+      only_paths: ["ofl/ibmplexmono"]
```
MikeMcQuaid, line 8: Could maybe even make this accept a string or an array?

Casks/font-ibm-plex-mono.rb
```diff
@@ -2,24 +2,25 @@
...
-  font "IBMPlexMono-Light.ttf"
-  font "IBMPlexMono-LightItalic.ttf"
-  font "IBMPlexMono-Medium.ttf"
-  font "IBMPlexMono-MediumItalic.ttf"
-  font "IBMPlexMono-Regular.ttf"
-  font "IBMPlexMono-SemiBold.ttf"
-  font "IBMPlexMono-SemiBoldItalic.ttf"
-  font "IBMPlexMono-Thin.ttf"
-  font "IBMPlexMono-ThinItalic.ttf"
+  font "ofl/ibmplexmono/IBMPlexMono-Bold.ttf"
```
MikeMcQuaid, line 12: I wonder if this bit could be stripped out given the `only_paths` above?

Casks/font-ibm-plex-mono.rb
MikeMcQuaid: > I tried removing branch, but it seems like we assume master is the default branch:

Ah, cool, all good, thanks!

MikeMcQuaid (approved):
Some pedantic comments but would happily see this ship as-is.

Casks/font-ibm-plex-mono.rb
hmarr: I did a little digging and it looks like that'd require a more significant change in the Cask code.

Right now the file is [moved](https://github.com/homebrew/brew/blob/6fc116318e53cc8dae1460a0896026399a75e383/Library/Homebrew/cask/artifact/moved.rb#L59) from a [single source location](https://github.com/homebrew/brew/blob/6fc116318e53cc8dae1460a0896026399a75e383/Library/Homebrew/cask/artifact/relocated.rb#L53), so we'd need to update that code to check for the presence of the file in multiple places. If we go down that route, there's also the possibility of collisions, so we'd have to say something like the first match wins, and we start by assuming it's relative to the root rather than one of the `only_paths` paths.

I think you don't need the path prefix for the SVN strategy because it only fetches a single subdirectory within the repository, so there's only one place to check (and the root of the checkout is the subdirectory, so you get that behaviour for free).

I'm happy to take this on if you think the improvement to the API would be worth it. Just to throw it out there: we could also change `only_paths` to `only_path`, and mimic the SVN behaviour. I suspect the 90+% usage for this feature will be a single path.

Casks/font-ibm-plex-mono.rb
hmarr: Yeah I like that. In the comment below I also mentioned the option of starting with _only_ a string to get rid of the complexity around removing the path prefix.

Casks/font-ibm-plex-mono.rb
MikeMcQuaid: > I'm happy to take this on if you think the improvement to the API would be worth it. Just to throw it out there: we could also change only_paths to only_path, and mimic the SVN behaviour. I suspect the 90+% usage for this feature will be a single path.

Yeh, I think it'd be worth it. I also like the idea of optimising for the 90% case and another method/DSL can be added if needed for other use-cases.

Thanks again @hmarr!

hmarr: I took this approach in https://github.com/Homebrew/brew/pull/14035 and it is indeed way neater – the diff of this PR has shrunk a lot as a result 🙂

hmarr:
Now that https://github.com/Homebrew/brew/pull/14035 has landed we should be able to start migrating fonts from svn to git 🎉 

This change will depend on the next release of brew (after 3.6.7) being installed. I _think_ that'll generally happen by `brew update --auto-update` running when `brew install` and friends are run. Are we fine to just wait for that release to come out to start updating this cask repo or is there a waiting period for `HOMEBREW_NO_AUTO_UPDATE` stragglers?

In terms of rollout, I propose:
1. We start with this font (`ibm-plex-mono`) as it has [~80 downloads per month](https://formulae.brew.sh/analytics/cask-install/30d/), which seems like enough that we'll find out if anything breaks, but no so many that a breakage would be disastrous.
2. Then if all goes well I'll bulk-update the other svn fonts in a follow-up PR.

bevanjkay:
Just added a "do not merge" tag so that we don't merge this before Brew 3.6.7 drops - CI should pass now.

MikeMcQuaid:
> In terms of rollout, I propose:
> 
> 1. We start with this font (`ibm-plex-mono`) as it has [~80 downloads per month](https://formulae.brew.sh/analytics/cask-install/30d/), which seems like enough that we'll find out if anything breaks, but no so many that a breakage would be disastrous.
> 2. Then if all goes well I'll bulk-update the other svn fonts in a follow-up PR.

Sounds good 👍🏻 



> Just added a "do not merge" tag so that we don't merge this before Brew 3.6.7 drops - CI should pass now.

Thanks @bevanjkay!

MikeMcQuaid (approved):

MikeMcQuaid:
3.6.8 was tagged so this can be merged. Thanks again @hmarr!

PR #7048, opened 2023-01-30
Update sha256 for font-iansui

vincenthsu:
The author replace the file causing hash mismatch.


p-linnane (approved):
Thank you @vincenthsu.

PR #7079, opened 2023-02-05
Remove appcasts.

reitermarkus:


miccal:
Besides `font-aileron`, the remaining ones download Windows executables?

bevanjkay:
@miccal Looks like there's a system in the Install methods to unpack the `.exe` files - so they're installing correctly - just failing out due to no having livechecks.

reitermarkus:
Yeah, the `.exe` are basically just executable archives extracted with `cabextract` or `7z` I think.

bevanjkay (approved):

PR #7336, opened 2023-04-03
a-f: Add trailing "/" to verified URL stanza

issyl0:
- Companion PR to https://github.com/Homebrew/brew/pull/15125.
- Splitting up the 430 files changed PR into batches.
- This does fonts from A to F.

-----

### DIFF
--- a/Casks/font-cantarell.rb
+++ b/Casks/font-cantarell.rb
@@ -3,7 +3,7 @@ cask "font-cantarell" do
   sha256 :no_check
 
   url "https://github.com/google/fonts.git",
-      verified:  "github.com/google/fonts",
+      verified:  "github.com/google/fonts/",
       branch:    "main",
       only_path: "ofl/cantarell"
   name "Cantarell"
--- a/Casks/font-castoro.rb
+++ b/Casks/font-castoro.rb
@@ -3,7 +3,7 @@ cask "font-castoro" do
   sha256 :no_check
 
   url "https://github.com/google/fonts.git",
-      verified:  "github.com/google/fonts",
+      verified:  "github.com/google/fonts/",
       branch:    "main",
       only_path: "ofl/castoro"
   name "Castoro"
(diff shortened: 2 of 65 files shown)


Casks/font-abeezee.rb
```diff
@@ -3,7 +3,7 @@
   sha256 :no_check
 
   url "https://github.com/google/fonts.git",
-      verified:  "github.com/google/fonts",
+      verified:  "github.com/google/fonts/",
```
reitermarkus, line 6: This is wrong.

issyl0: Hmm well that means that all of them are wrong then because it's the same change 400 times over. What a waste of time. 😩

issyl0: Could you explain why? Maybe I don't understand the purpose of this verified stanza?

reitermarkus: The URL `https://github.com/google/fonts.git` does not contain `github.com/google/fonts/`, so the audit fails.

I guess having `github.com/google/` verified would be enough since it verifies the GitHub org which controls the `fonts` repo.

It is meant to make sure that a maintainer has verified the URL is correct, if it's not on the same domain as the homepage.

In this repo, we will also need to change the workflow which auto-updates Google Fonts if we change this.

issyl0: Is the audit wrong? Is the `url` wrong (does it need the `.git`)? Is the `verified` wrong (well obviously yes but I'm wondering which thing should be fixed here)?

reitermarkus: The URL is correct. If it was just `https://github.com/google/fonts` I think we would need to explicitly specify `using: :git` for all of these.

The audit is also correct, otherwise the audit would need a special case for GitHub URLs.

The cop should only flag this if the URL also contains a slash.

PR #7409, opened 2023-04-21
Added Monocraft Nerd

e-minguez:
The Monocraft font has also a patched Nerd version which it would be nice to have as well.


reitermarkus:
I think this should be named `font-monocraft-nerd-font` to be in line with the other nerd fonts.

miccal (approved):

miccal:
Thank you @e-minguez.

PR #7771, opened 2023-06-13
cask_renames: rename font-source-sans-pro and font-source-serif-pro

bevanjkay:
`font-source-sans-pro` and `font-source-sans-serif` were removed by the PR https://github.com/Homebrew/homebrew-cask-fonts/pull/7742 as they have been replaced upstream by variable font versions but were not removed from the `google/fonts` repo until https://github.com/google/fonts/pull/6328


Bo98:
Technically it's a different font name so uninstalling the old one could break existing documents.

PR #8096, opened 2023-08-18
workflows/automerge: add action

p-linnane:
Adding to match the main cask repo.


bevanjkay:
@p-linnane Perhaps we should add this to the workflow syncing system? It doesn't look like there's any repo specific logic.

p-linnane:
@bevanjkay We can give it a shot.

p-linnane:
Closing in favor of #8119.

PR #8643, opened 2023-11-05
B612 version and conflict

recursiveribbons:


bevanjkay:
I think we should remove the mono variants from the non-mono cask and remove the conflicts with. This would allow both to be installed at once. 

Most fonts in `homebrew-cask-fonts` separate mono to another cask.

recursiveribbons:
@bevanjkay We could also remove the mono cask entirely and have the normal cask install both?

bevanjkay:
Google Fonts are managed systematically so best to not include the mono variant in the b6512 cask.

bevanjkay (approved):

PR #8720, opened 2023-11-28
font-noto-serif-cjk 2.002

muzimuzhi:
**Why need an update**

The old url https://noto-website-2.storage.googleapis.com/pkgs/NotoSerifCJK.ttc.zip provides an outdated version 1.001, packed back in 2017. While the latest version is 2.002, see https://github.com/notofonts/noto-cjk/releases/tag/Serif2.002.

**Why updated in this way**

https://fonts.google.com only provides variable ttf font file in Noto Serif CJK bundle for each language, for example `NotoSerifSC[wght].ttf` and its corresponding cask [`font-noto-serif-sc.rb`](https://github.com/Homebrew/homebrew-cask-fonts/blob/master/Casks/font-noto-serif-sc.rb) for Noto Serif CJK SC (Simplified Chinese). But this cask provides a static super ttc font file, which is only available from releases of repository https://github.com/notofonts/noto-cjk.

**Remark**

More casks need to be updated in this way, including symmetric Noto Sans CJK cask [`font-noto-sans-cjk.rb`](https://github.com/Homebrew/homebrew-cask-fonts/blob/master/Casks/font-noto-sans-cjk.rb) and language specific Serif and Sans casks like [`font-noto-sans-cjk-sc.rb`](https://github.com/Homebrew/homebrew-cask-fonts/blob/master/Casks/font-noto-sans-cjk-sc.rb).

I started with only one cask to check I did everything right.

**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making any changes to a cask, existing or new, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [ ] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=Issues).
- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
[...]

PR #8721, opened 2023-11-29
Bump font-noto-serif-cjk-(jk|kr|sc|tc) 2.002

muzimuzhi:
**Important:** *Do not tick a checkbox if you haven’t performed its action.* Honesty is indispensable for a smooth review process.

_In the following questions `<cask>` is the token of the cask you're submitting._

After making any changes to a cask, existing or new, verify:

- [x] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [x] `brew audit --cask --online <cask>` is error-free.
- [x] `brew style --fix <cask>` reports no offenses.

Additionally, **if adding a new cask**:

- [ ] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=Issues).
- [ ] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
- [ ] `brew audit --new-cask <cask>` worked successfully.
- [ ] `brew install --cask <cask>` worked successfully.
- [ ] `brew uninstall --cask <cask>` worked successfully.

PR #8732, opened 2023-12-05
font-noto-sans-cjk 2.004

muzimuzhi:
Symmetric PR of #8720.

The language specific casks for Noto Sans CJK are a bit different than their Serif correspondings, hence will be updated in a separate PR.

### DIFF
--- a/Casks/font-noto-sans-cjk.rb
+++ b/Casks/font-noto-sans-cjk.rb
@@ -1,11 +1,16 @@
 cask "font-noto-sans-cjk" do
-  version :latest
-  sha256 :no_check
+  version "2.004"
+  sha256 "a56077736038875f3a5b1e6715b2713a2bd049cc0bf0dfdf645927b070d38bfc"
 
-  url "https://noto-website-2.storage.googleapis.com/pkgs/NotoSansCJK.ttc.zip",
-      verified: "noto-website-2.storage.googleapis.com/"
+  url "https://github.com/notofonts/noto-cjk/releases/download/Sans#{version}/00_NotoSansCJK.ttc.zip"
   name "Noto Sans CJK"
-  homepage "https://www.google.com/get/noto/help/cjk/"
+  desc "Static Super OTC"
+  homepage "https://github.com/notofonts/noto-cjk/tree/main/Sans"
+
+  livecheck do
+    url :url
+    regex(/^Sans(\d+(?:\.\d+)+)$/i)
+  end
 
   font "NotoSansCJK.ttc"
 
--- a/Casks/font-noto-sans-cjk.rb
+++ b/Casks/font-noto-sans-cjk.rb
@@ -9,7 +9,8 @@ cask "font-noto-sans-cjk" do
 
   livecheck do
     url :url
-    regex(/^Sans(\d+(?:\.\d+)+)$/i)
+    regex(/^(?:Noto)?Sans[._-]?v?(\d+(?:\.\d+)+)$/i)
+    strategy :github_releases
   end
 
   font "NotoSansCJK.ttc"


muzimuzhi:
I took "**Do** start with an uppercase letter." in https://docs.brew.sh/Cask-Cookbook#dos-and-donts the other way around. Sorry.

Casks/font-noto-sans-cjk.rb
```diff
@@ -1,11 +1,16 @@
...
-      verified: "noto-website-2.storage.googleapis.com/"
+  url "https://github.com/notofonts/noto-cjk/releases/download/Sans#{version}/00_NotoSansCJK.ttc.zip"
   name "Noto Sans CJK"
-  homepage "https://www.google.com/get/noto/help/cjk/"
+  desc "Static Super OTC"
+  homepage "https://github.com/notofonts/noto-cjk/tree/main/Sans"
+
+  livecheck do
+    url :url
+    regex(/^Sans(\d+(?:\.\d+)+)$/i)
```
samford, line 12:
```suggestion
    regex(/^Sans[._-]?v?(\d+(?:\.\d+)+)$/i)
    strategy :github_releases
```

The cask `url` is a GitHub release asset, so this should use either `GithubLatest` or `GithubReleases`. Upstream creates alternating Serif and Sans releases and the "latest" release may be either, so we have to use `GithubReleases` here.

Past that, this suggestion adds some optional components to the regex, as there's some variety in their past tag format. Hopefully they will stick with the `Sans2.004` format going forward, though.

muzimuzhi: > Upstream creates alternating Serif and Sans releases and the "latest" release may be either, so we have to use `GithubReleases` here.

Ah thanks! Then I guess all the `font-noto-serif-cjk-*` casks need similar strategy. Do they?



> Past that, this suggestion adds some optional components to the regex, as there's some variety in their past tag format.

You meant tags in other repositories of notofonts organisation? Because it seems for `notofonts/noto-cjk` the variety in tag formats is not to be matched by the updated regex:

```
$ gh release list --repo notofonts/noto-cjk

TITLE                                           TYPE    TAG NAME        PUBLISHED
Noto Serif CJK Version 2.002 (OTF, OTC, Sup...  Latest  Serif2.002      about 3 months ago
Noto Sans CJK Version 2.004 (OTF, OTC, Supe...          Sans2.004       about 1 year ago
Noto Serif CJK Version Serif2.001 (OTF, OTC...          Serif2.001      about 1 year ago
Noto CJK v20201206 as of 2020.12.06                     v20201206-cjk   about 2 years ago
Noto Sans CJK V2.001                                    NotoSansV2.001  about 4 years ago
Match Adobe's 1.004 release that corrects v...          v1.004          about 8 years ago
```

samford: > Then I guess all the `font-noto-serif-cjk-*` casks need similar strategy

Since the files in the variant casks are provided as part of the same release as `font-noto-sans-cjk`, we can simply use `cask "font-noto-sans-cjk"` in those `livecheck` blocks to avoid the duplication. [We only use `#cask` when there's a clear primary/secondary relationship between the casks. In this cask `font-noto-sans-cjk` is the primary cask and these are secondary variants, so it makes sense here.]

I'll push a commit to that branch in a moment to update those `livecheck` blocks but we'll need to rebase that PR after this one is merged to use the `livecheck` block from this PR.

> You meant tags in other repositories of notofonts organisation? Because it seems for `notofonts/noto-cjk` the variety in tag formats is not to be matched by the updated regex

I was referring to the tags in the notofonts/noto-cjk repository, as you've listed. `/^Sans[._-]?v?(\d+(?:\.\d+)+)$/i` will match tags like `Sans2.004`, `SansV2.004`, `Sans-v2.004`, etc. I deliberately omitted an optional leading `(?:Noto)?` for now, so it won't match the older `NotoSansV2.001` (which doesn't have release assets). That said, there's no harm in matching the older release (with 2.004 being newer than it). I'll update the suggestion above.

muzimuzhi: > we can simply use `cask "font-noto-sans-cjk"` in those `livecheck` blocks to avoid the duplication.

Thank you again! Now I know something new about `livecheck`.
https://docs.brew.sh/Brew-Livecheck#referenced-formulacask.

> I've pushed a commit to that branch to update those `livecheck` blocks but we'll need to rebase that PR after this one is merged.

I'll open another PR updating `livecheck` for Noto Serif CJK cask and its variants (serif, not sans), which were updated previously in #8720 and #8721.

bevanjkay (approved):

PR #8794, opened 2023-12-17
Use `deprecate!` instead of `discontinued`

Rylan12:
See https://github.com/Homebrew/brew/pull/16292

Move all instances of `discontinued` in the `caveats` stanza to a `deprecate!` call.

For now, I've left the date for all of these as today (`2023-12-17`) because it doesn't seem to me like it will be a good use of my time to go through each and figure out an appropriate date. If that isn't the best solution, though, I'm happy to hear about alternatives.

This PR was made using `brew style --fix` with https://github.com/Homebrew/brew/pull/16351 checked out, and then the non-autocorrectable (when the `caveats` stanza includes more than just `discontinued`) offenses were handled manually.

Note: this PR cannot be merged until a new Homebrew/brew tag (4.2.0) is made.

### DIFF
--- a/Casks/font-amiri.rb
+++ b/Casks/font-amiri.rb
@@ -8,16 +8,14 @@ cask "font-amiri" do
   desc "Classical Arabic typeface in Naskh style"
   homepage "https://www.amirifont.org/"
 
+  deprecate! date: "2023-12-17", because: :discontinued
+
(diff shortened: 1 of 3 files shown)


p-linnane (approved):
Thanks! Great work!

PR #8984, opened 2024-01-30
Add variable noto fonts

sharpenedblade:

Livecheck is throwing an error, but the version is actually up to date.

- [x] Named the cask according to the [token reference](https://docs.brew.sh/Cask-Cookbook#token-reference).
- [x] Checked the cask is submitted to [the correct repo](https://docs.brew.sh/Acceptable-Casks#finding-a-home-for-your-cask).
It says the repo is not popular enough. This font (from the same source repo) is already packaged here as a non-variable font so I dont think this applies here. The github repo is just a download link, the actual font is much more popular.
- [x] `HOMEBREW_NO_INSTALL_FROM_API=1 brew install --cask <cask>` worked successfully.
- [x] `brew uninstall --cask <cask>` worked successfully.


bevanjkay:
Thanks for the submission @sharpenedblade. Is there any reason these versions should not be distributed as the current `font-noto-sans` etc...?

sharpenedblade:
> Thanks for the submission @sharpenedblade. Is there any reason these versions should not be used as the current `font-noto-sans` etc...?

I dont think so, but some older apps dont handle variable fonts right, and I noticed that other fonts were packaged in the same way.

bevanjkay:
If this is the current official distribution of the fonts then yes. 

In other cases it is likely because the variable fonts are packaged separately upstream, so cannot be installed from the same archive (therefore requiring a separate cask).

sharpenedblade:
> In other cases it is likely because the variable fonts are packaged separately upstream, so cannot be installed from the same archive (therefore requiring a separate cask).

Having both regular and variable fonts installed causes inconsistant rendering in a lot of apps.

bevanjkay:
According to the documentation, we prefer variable fonts over static fonts, so this should be ok, unless it causes issues in up to date applications.
https://github.com/Homebrew/homebrew-cask-fonts/blob/master/CONTRIBUTING.md#multiple-font-formats

sharpenedblade:
> According to the documentation, we prefer variable fonts over static fonts, so this should be ok, unless it causes issues in up to date applications. https://github.com/Homebrew/homebrew-cask-fonts/blob/master/CONTRIBUTING.md#multiple-font-formats

I updated the PR

bevanjkay:
Tests are currently failing. But in principle I'm happy with the changes.

sharpenedblade:
> Tests are currently failing. But in principle I'm happy with the changes.

Fixed, the only error is about popularity but there is nothing I can do about that.

Casks/font-noto-sans-mono.rb
```diff
@@ -1,48 +1,14 @@
 cask "font-noto-sans-mono" do
-  version :latest
+  version "2.014"
   sha256 :no_check
```
bevanjkay, line 3: Now that we are using a versioned `url` a checksum can be added to each cask

bevanjkay (changes requested):

Casks/font-noto-sans-mono.rb
bevanjkay: Should be good to go after that change

sharpenedblade: I added checksums

sharpenedblade:
All tests are passing now

Casks/font-noto-sans-mono.rb
```diff
@@ -1,48 +1,19 @@
...
-  font "NotoSansMono-SemiCondensedBold.ttf"
-  font "NotoSansMono-SemiCondensedExtraBold.ttf"
-  font "NotoSansMono-SemiCondensedExtraLight.ttf"
-  font "NotoSansMono-SemiCondensedLight.ttf"
-  font "NotoSansMono-SemiCondensedMedium.ttf"
-  font "NotoSansMono-SemiCondensedSemiBold.ttf"
-  font "NotoSansMono-SemiCondensedThin.ttf"
-  font "NotoSansMono-Thin.ttf"
+  livecheck do
+    url "https://github.com/notofonts/latin-greek-cyrillic.git"
```
bevanjkay, line 12: all of these should be adjusted to

```suggestion
    url :url
```

Sorry for the back and forth @sharpenedblade

bevanjkay (changes requested):

Casks/font-noto-sans-mono.rb
sharpenedblade: Fixed

bevanjkay (approved):

bevanjkay:
Thank you @sharpenedblade

PR #9018, opened 2024-02-08
font-meslo-for-powerlevel10k: use official source

bevanjkay:

----

Continuation of #8988 to use the official source.
The downside here is that it won't detect when there is an update (but will replace local files with upstream, if `brew upgrade --greedy` is ran), and the whole git repository has to be checked out due to the files not being located within a subfolder.

CC: @romkatv

### DIFF
--- a/Casks/font-meslo-for-powerlevel10k.rb
+++ b/Casks/font-meslo-for-powerlevel10k.rb
@@ -1,11 +1,11 @@
 cask "font-meslo-for-powerlevel10k" do
-  version "1.2.1"
-  sha256 "2581fcaa4c2dd9bb4cbb5cf6df318bfce925b51a130efad5e0d04fd68c46f4e6"
+  version "2.3.3"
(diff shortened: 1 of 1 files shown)


romkatv:
> The downside here is that it won't detect when there is an update (but will replace local files with upstream, if `brew upgrade --greedy` is ran), and the whole git repository has to be checked out due to the files not being located within a subfolder.

What is the right way to set it up without the downsides you mentioned? If I create another repo just for the font files, would that help?

bevanjkay:
@romkatv The best way to be able to notify users when a new update is available for the font would be a repo that has a new release (and therefore also archive) on each update. Then the caskfile could contain a version. 

If this is too difficult having the files in their own repo would be less download overhead, or even keeping them in the same repo but in a subfolder would save some bandwidth.

romkatv:
How about this? https://github.com/romkatv/powerlevel10k-media/releases/tag/v2.3.3

p-linnane:
Thanks for the help here @romkatv! This should be all we need.

p-linnane (approved):

romkatv:
Thanks a bunch!

PR #9042, opened 2024-02-22
font-linux-libertine: update version, livecheck

samford:

-----

The `livecheck` block for `font-linux-libertine` and `font-linux-biolinum` uses a SourceForge RSS feed URL that's the same as the default URL from the `Sourceforge` strategy. In the process of addressing this redundancy, I tested whether the `livecheck` block was required and saw that it's necessary because the 5.0.0 release tarball only uses a date, so `2011_05_22` would erroneously be returned as newest. Looking through releases, there are some with only a version (e.g., 4.4.1), some with a version that has a numeric suffix (e.g., 4.7.5-2), some with a version and date (e.g., 5.3.0_2012_07_02), and the aforementioned one with only a date.

The `font-linux-biolinum` and `font-linux-libertine` casks both use a `5.3.0_2012_07_02` version. This is what's used in the tarball file name but the upstream [version directories](https://sourceforge.net/projects/linuxlibertine/files/linuxlibertine/) are like `1.2.3` or `1.2.3-4`. Considering this along with the variety of version formats above, I think it makes sense to use a CSV version format like `5.3.0,2012_07_02` instead (along with `version.tr(",", "_")` in the `url`).

This PR updates the `version`/`url` and the `livecheck` block for these casks accordingly.

### DIFF
--- a/Casks/font-linux-biolinum.rb
+++ b/Casks/font-linux-biolinum.rb
@@ -1,15 +1,24 @@
 cask "font-linux-biolinum" do
-  version "5.3.0_2012_07_02"
+  version "5.3.0,2012_07_02"
   sha256 "24a593a949808d976850131a953c0c0d7a72299531dfbb348191964cc038d75d"
 
-  url "https://downloads.sourceforge.net/linuxlibertine/LinLibertineTTF_#{version}.tgz",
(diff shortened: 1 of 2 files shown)


miccal:
Thanks @samford.

miccal (approved):

PR #9044, opened 2024-02-24
Update Sourceforge livecheck blocks

samford:

-----

The casks in this PR contain a `livecheck` block where the specified `url` string is the same as the default URL generated by livecheck's `Sourceforge` strategy (i.e., they're redundant). This updates the URLs to avoid this redundancy. In this case, I've added a `path` query string parameter to each, to restrict the RSS feed to the directory we're interested in.

This PR also:

* Updates the `regex` for `font-computer-modern`, `font-han-nom-a`, and `font-junicode` to improve them and bring them more in line with typical `Sourceforge` regexes.
* Adds a `strategy` block for `junicode`, to convert the older `0-7-8` version format in file names to `0.7.8.
* Removes the `regex` from `font-dejavu`, as the default from the `Sourceforge` strategy accomplishes the same thing. [We could technically use something like `regex(%r{url=.*?/dejavu-fonts-ttf[._-]v?(\d+(?:\.\d+)+)\.}i)` if we strictly want to match `dejavu-fonts-ttf` files but that could quietly fail to match a new version if they change the file name format. Granted, it would also match a version that doesn't include the file we need but that doesn't seem to be an issue based on past versions.]

Edit: I added a `desc` for `font-computer-modern` but feel free to update it.

### DIFF
--- a/Casks/font-computer-modern.rb
+++ b/Casks/font-computer-modern.rb
@@ -5,11 +5,12 @@ cask "font-computer-modern" do
   url "https://downloads.sourceforge.net/cm-unicode/cm-unicode-#{version}-ttf.tar.xz",
       verified: "downloads.sourceforge.net/cm-unicode/"
   name "Computer Modern"
+  desc "Multilingual unicode fonts, mainly for X applications"
   homepage "https://cm-unicode.sourceforge.io/"
 
   livecheck do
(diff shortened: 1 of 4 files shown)


p-linnane (approved):
Awesome, thanks!

PR #9059, opened 2024-03-04
Fix `font-noto-sans`

wenkokke:

- [X] The submission is for [a stable version](https://docs.brew.sh/Acceptable-Casks#stable-versions) or [documented exception](https://docs.brew.sh/Acceptable-Casks#but-there-is-no-stable-version).
- [X] `brew audit --cask --online <cask>` is error-free.
- [X] `brew style --fix <cask>` reports no offenses.

---

This PR fixes issue #9060.


bevanjkay:
If the variable font files can be installed alongside the static files then they should be retained.

wenkokke:
So it should be both the original font statements and the ones I added?

p-linnane:
@wenkokke Please address @bevanjkay's feedback so we can gets this merged. Otherwise it will be closed. Thanks.

p-linnane (approved):

p-linnane:
Thanks @wenkokke!

PR #9359, opened 2024-05-08
Document installation via Brewfile

jasonkarns:
Font installation has changed a few times (particularly from caskroom tap) and googling for how to install fonts via Brewfile surfaces a few conflicting approaches.

I think it would be beneficial if taps documented how they were to be used within a Brewfile. This is especially helpful for homebrew's first-party taps, to remove any potential confusion that there may perhaps be a `font "inconsolata"` mechanism or similar; given that `brew`, `cask`, `mas`, `whalebrew` and `vscode` are all first-class Brewfile methods. It also eliminates any confusion about whether the `font-` prefix should be present in the Brewfile, given that there are other prefixes in the homebrew ecosystem that can conventionally be omitted in certain cases. Thus, an explicit example seems worthwhile.

---
_(not a cask PR, so the following is not applicable)_

- [ ] Checked the cask was not [already refused](https://github.com/Homebrew/homebrew-cask-fonts/search?q=is%3Aclosed&type=pullrequests).

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,13 @@ $ brew tap homebrew/cask-fonts         # You only need to do this once!
 $ brew install font-inconsolata
 ```
 
+Or in a [brew-bundle](https://github.com/Homebrew/homebrew-bundle) `Brewfile`:
+
+```ruby
+tap "homebrew/cask-fonts"
+cask "font-inconsolata"
+```
+
 ## Submitting a Cask to this repository
 
 See [CONTRIBUTING.md](CONTRIBUTING.md).


bevanjkay:
Thanks for suggestion this. I think this is more suitable to be added to the[ `homebrew-bundle`](github.com/homebrew/homebrew-bundle) readme.

It's also worth noting that the plan is to migrate everything in this repo to the `homebrew-cask` repo, and this repo would no longer be tapped.

jasonkarns:
> I think this is more suitable to be added to the[ homebrew-bundle](https://github.com/Homebrew/homebrew-cask-fonts/pull/github.com/homebrew/homebrew-bundle) readme.

Maybe? It would be fine for homebrew-bundle to exhaustively document installation of any formulae for core taps. However, it's not feasible to document an example for _all_ taps. Which is why I think this kind of documentation belongs in the taps themselves (whether core or 3rd party). Indeed, when I was confused about how to install this, the first place I looked is the tap, not brew-bundle. (I wouldn't expect brew-bundle to have an example for arbitrary taps, whereas a tap could reasonably document the Brewfile installation alternative.)

For this reason, I suggested (and it has landed) the tap template including a brewfile example: https://github.com/Homebrew/brew/pull/17256

> the plan is to migrate everything

If this tap is not long for this world, we certainly don't need to bother modifying the readme. But if there's a chance this merge doesn't take place soon-ish (your call on what "soon" means :D ), I would think it's still valuable to bring the tap into consistency with the tap template for uniformity if nothing else?

bevanjkay:
Thanks @jasonkarns for your understanding. The hope is that this repo will be migrated within the next week or two.

